PR #17, opened 2017-12-11
Add option to set toplevel folders via parameters from the outside world

### LINKED ISSUE #16: Only check modified addons in a PR run
It isn't useful to check all addons as one has to look at the log if the reason for a failing check is caused by the changed addon or by some other addon.

Razzeee:
@Rechi 

Closes #16

### DIFF
--- a/check_repo.py
+++ b/check_repo.py
@@ -19,7 +19,12 @@ def check_repo():
         os.path.dirname(os.path.realpath(__file__)), os.pardir))
     print("Repo path " + repo_path)
 
-    toplevel_folders = sorted(next(os.walk(repo_path))[1])
+    parameters = sys.argv[1:]
+    if len(parameters) == 0:
+        toplevel_folders = sorted(next(os.walk(repo_path))[1])
+    else:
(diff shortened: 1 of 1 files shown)


Razzeee:
If you edit the travis script, it might be nice to always build master without parameters so that everything get's build and only build pr's with the git diff parameters.

check_repo.py
```diff
@@ -19,7 +19,12 @@ def check_repo():
         os.path.dirname(os.path.realpath(__file__)), os.pardir))
     print("Repo path " + repo_path)
 
-    toplevel_folders = sorted(next(os.walk(repo_path))[1])
+    parameters = sys.argv[1:]
+    if len(parameters) == 0:
+        toplevel_folders = sorted(next(os.walk(repo_path))[1])
+    else:
+        toplevel_folders = sorted(parameters)
```
Rechi: maybe build the intersection of actuall toplevel_folders and parameters to be more fail safe

Razzeee: Talked with rechi and this is not necessary

PR #26, opened 2017-12-28
Add check for transparency

Razzeee:
Ping @romanvm as I hope you have some Pillow knowledge. It should be fine, but needs a double check.


romanvm:
Sorry, I did not reply to you earlier, but I'm not very familiar with PIL/Pillow.

Razzeee:
Well if more png variations pop up we will probably figure it out then :)

PR #32, opened 2018-02-08
[travis] exclude deleted addons from PR check

Rechi:


Rechi:
pushed the fix to all repos which already had a ` .travis` file

Issue #27, opened 2017-12-28
Auto comment found issues on the PR


MartijnKaijser:
When there are issues found it would be nice they would be automatically added as comment on the PR. If need accompanied with url to some helpful page.

Razzeee:
As bulk as one comment or split and on affected line of the file? Pretty different problem. Also we probably need to store our secret somewhere, will have to look if travis is able to do that for us. As the rest is open source.

This seems helpful for the secret stuff: https://damien.pobel.fr/post/github-api-from-travisci/
And this library is the most popular github api for python: https://github.com/PyGithub/PyGithub

MartijnKaijser:
Whatever is easiest which will probably be a bulk comment.

Razzeee:
Yes, spares us quite some work

Razzeee:
I just merged that 🥇

PR #34, opened 2018-02-11
Auto comment on issues found in pull request

mzfr:
It fixes issue #27 

This will help in informing  contributors about the problems in their add-ons.

I have tested it locally but haven't tested it on Travis-CI.

### DIFF
--- a/check_addon.py
+++ b/check_addon.py
@@ -8,6 +8,8 @@ from common import colorPrint, check_config, has_transparency
 
 REL_PATH = ""
 comments = []
+
+
 def _find_file(name, path):
     for file_name in os.listdir(path):
         match = re.match(name, file_name, re.IGNORECASE)
@@ -66,22 +68,27 @@ def start(error_counter, addon_path, config=None):
         if len(addon_xml.findall("*//broken")) == 0:
             file_index = _create_file_index(addon_path)
 
-            error_counter = _check_for_invalid_xml_files(error_counter, file_index)
+            error_counter = _check_for_invalid_xml_files(
+                error_counter, file_index)
 
             error_counter = _check_for_invalid_json_files(
                 error_counter, file_index)
 
-            error_counter = _check_artwork(error_counter, addon_path, addon_xml, file_index)
+            error_counter = _check_artwork(
+                error_counter, addon_path, addon_xml, file_index)
 
             if check_config(config, "check_license_file_exists"):
                 # check if license file is existing
-                error_counter = _addon_file_exists(error_counter, addon_path, "LICENSE\.txt|LICENSE\.md|LICENSE")
+                error_counter = _addon_file_exists(
+                    error_counter, addon_path, "LICENSE\.txt|LICENSE\.md|LICENSE")
 
             if check_config(config, "check_legacy_strings_xml"):
-                error_counter = _check_for_legacy_strings_xml(error_counter, addon_path)
+                error_counter = _check_for_legacy_strings_xml(
+                    error_counter, addon_path)
 
             if check_config(config, "check_legacy_language_path"):
-                error_counter = _check_for_legacy_language_path(error_counter, addon_path)
+                error_counter = _check_for_legacy_language_path(
+                    error_counter, addon_path)
 
             # Kodi 18 Leia + deprecations
             if check_config(config, "check_kodi_leia_deprecations"):
@@ -93,9 +100,11 @@ def start(error_counter, addon_path, config=None):
                     [], [".py", ".xml"])
 
             # General blacklist
-            error_counter = _find_blacklisted_strings(error_counter, addon_path, [], [], [])
+            error_counter = _find_blacklisted_strings(
+                error_counter, addon_path, [], [], [])
 
-            error_counter = _check_file_whitelist(error_counter, file_index, addon_path)
+            error_counter = _check_file_whitelist(
+                error_counter, file_index, addon_path)
         else:
             print("Addon marked as broken - skipping")
 
@@ -110,7 +119,8 @@ def _check_for_invalid_xml_files(error_counter, file_index):
                 # Just try if we can successfully parse it
                 xml.etree.ElementTree.parse(xml_path)
             except xml.etree.ElementTree.ParseError:
-                error_counter = _logProblem(error_counter, "Invalid xml found. %s" % relative_path(xml_path))
+                error_counter = _logProblem(
+                    error_counter, "Invalid xml found. %s" % relative_path(xml_path))
 
     return error_counter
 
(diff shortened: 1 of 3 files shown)


check_repo.py
```diff
@@ -33,15 +32,9 @@ def check_repo():
...
             error_counter = check_addon.start(error_counter, addon_path, config)
-
-    if error_counter["problems"] > 0:
-        colorPrint("We found %s problems and %s warnings, please check the logfile." % (
-            error_counter["problems"], error_counter["warnings"]), "31")
-        sys.exit(1)
-    elif error_counter["warnings"] > 0:
-        # If we only found warnings, don't mark the build as broken
-        colorPrint("We found %s problems and %s warnings, please check the logfile." % (
-            error_counter["problems"], error_counter["warnings"]), "35")
```
Rechi: this should be still kept
- without `sys.exit(1)` PRs won't be marked as failed
- travis log should still output an end result

git_comments.py
```diff
@@ -0,0 +1,34 @@
...
+import json
+import requests
+
+
+def comment_on_pull(errors):
+
+    # These are default Travis variables identifying a pull request
+    # https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables
+    repo = os.environ["TRAVIS_REPO_SLUG"]
+    pr = os.environ["TRAVIS_PULL_REQUEST"]
```
Rechi: script also runs for non PR builds so skip the comment if it is not a PR build

check_repo.py
Razzeee: And as an info, this is designed to be able to run on PRs/repos via travis and on your local machine, without modification. So it might make sense to make the github comments opt-in via the `.tests-config.json`?

check_addon.py
```diff
@@ -302,11 +302,12 @@ def _check_file_whitelist(error_counter, file_index, addon_path):
 
 def _logProblem(error_counter, problem_string):
     colorPrint("PROBLEM: %s" % problem_string, "31")
+    comments.append("PROBLEM: %s" % problem_string)
```
Razzeee: Might be nice have a combined api for your comments and the colorPrint interface. As the message should always be identical.

Razzeee:
Thank you for this, looks very promissing. I will try to run it on my github branch via travis as soon as you are able to fix the minors :)

mzfr:
I've fixed the issues and also tested on travis

check_repo.py
```diff
@@ -33,6 +32,9 @@ def check_repo():
         if addon_folder[0] != '.':
             addon_path = os.path.join(repo_path, addon_folder)
             error_counter = check_addon.start(error_counter, addon_path, config)
+    
+    if check_addon.comments:
```
Razzeee: This means that it will only comment when it find problems? Might be nice to always comment with everything is fine with your commit?
Not sure, @MartijnKaijser for spam analysis ;)

Razzeee: This would be the line were you load the config file instead and default to false ;)

MartijnKaijser: When all is fine it already shows the green check mark ;)
Option perhaps would be to use labels like "checks passed" for this instead of a comment? If it fails also set "Checks failed" label.
Of course this also needs to unset/set these when situation changes. This allows to filter the open PRs.

mzfr:
Made few changes suggested by @MartijnKaijser and @Razzeee

mzfr:
@Razzeee I have added option for ```.test-config.json```.

mzfr:
@razzeee Are there any other changes required in this ?

Razzeee:
I'm a bit short on review time at the moment. The only thing you could do is sqash your commits if you know how to do that :)

Razzeee:
I think the travis ci file also needs changes as you added some dependencies that are not installed on ci otherwise

Razzeee:
I took the liberty to add `requests` in the `travis-ci.yml` examples

Razzeee:
It's not perfect yet, check https://github.com/Razzeee/repo-plugins/pull/1 for an example.

I'm not sure if it might be prettier to show the errors in a markdown list? Might get very long, but at least you don't need to scroll horizontally.

mzfr:
@Razzeee  I have made changes suggested by you.

check_repo.py
```diff
@@ -33,7 +32,18 @@ def check_repo():
...
+
+    if check_addon.check_config(config, "comment_on_pull"):
+        if check_addon.comments:
+            GithubAPI().comment_on_pull(check_addon.comments)
+            GithubAPI().set_label(["Checks failed"])
+        else:
+            GithubAPI().remove_label(["Checks failed"])
+            GithubAPI().set_label(["Checks passed"])
+    else:
+        colorPrint("No config file found", "7")
```
Razzeee: This might not mean that the config file is not found, it might mean that it's just set to false

Razzeee:
Tested a bit more and it's failing quiet often due to

```
{'message': "API rate limit exceeded for 52.45.185.117. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)", 'documentation_url': 'https://developer.github.com/v3/#rate-limiting'}
```

check_repo.py
mzfr: I think I should remove this because if someone is intentionally going to set config to false then we don't have to show them any message saying "Config is set to false".

mzfr:
@Razzeee I have made the changes so now API Rate limit have increased from 60 to 5000(It's the [maximum](https://developer.github.com/v3/#rate-limiting)).

Razzeee:
Seems awy more relyable 🎉 

But formatting is still off https://github.com/Razzeee/repo-plugins/pull/1#issuecomment-369303762

check_addon.py
```diff
@@ -320,11 +339,13 @@ def relative_path(file_path):
...
```
Razzeee: Maybe we can have two lists, comments_problem and problems_warning. Then we could omit the PROBLEM: and WARNING: on each line and just write it one time followed by the respective comment list.

mzfr:
@Razzeee I have made changes suggested by you

Razzeee:
This is perfect, thank you 🌮 

And sorry for taking so long to review and having you go back so many times 💃

Razzeee:
Crap, would have been to easy.
Just tried to set it up and it keeps failing due to the `GITHUB_TOKEN` enviroment variable not being set anymore.
https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions

mzfr:
@Razzeee Wow, this is bad. There must be some recourse against this, right? Considering so many other tools do this GitHub comments thing. 

At worst, we would have to stop using Travis CI and maybe dedicate our own server for this, which would work via GitHub events hook API. We could try to use [existing tools](https://github.com/bluzi/travis-buddy) for this or try to come up with our own lean implementation in Python. for. eg. [this is](https://github.com/VSCodeVim/Vim/pull/2350#issuecomment-362044605) how TravisBuddy's comments look, notice that they are on a pull created from a fork of the repo.

But perhaps there is a better way to deal with this?

MartijnKaijser:
I think we'll have to add some github hook which we can grant access to be able to comment

Razzeee:
Well that travis buddy is running on their own infrastructure which obviously can keep the api key secret. While we try to let travis do that for us :/

PR #35, opened 2018-02-14
Add schema validation

### LINKED ISSUE #33: Verify addon.xml against a xml schema
I don't think we have a schema but we could create one.

mzfr:
Fixes #33 

I have tested many addon.xml files against the schema's but none of them matches.I have taken schema's definitions from [Here](https://github.com/xbmc/xbmc/tree/master/addons)

I am not sure if this is the correct way of doing it or not.I think we'll have to create a new schema for addon.xml but again I am not sure.

Let me know about all the changes that are required.

### DIFF
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -16,6 +16,7 @@ from . import check_files
 from . import check_string
 from . import check_py3_compatibility
 from . import common
+from . import schema_validation
 
 ROOT_URL = "http://mirrors.kodi.tv/addons/{branch}/addons.xml.gz"
 LOGGER = logging.getLogger(__name__)
@@ -39,6 +40,8 @@ def start(addon_path, branch_name, all_repo_addons, pr, config=None):
     if addon_xml is not None:
         if len(addon_xml.findall("*//broken")) == 0:
             file_index = handle_files.create_file_index(addon_path)
+
+            schema_validation.schemas(addon_report, parsed_xml, branch_name)
 
             check_dependencies.check_addon_dependencies(addon_report, repo_addons, parsed_xml, branch_name)
 
--- a/kodi_addon_checker/schema_validation.py
+++ b/kodi_addon_checker/schema_validation.py
@@ -1,0 +1,116 @@
+from os.path import abspath, dirname, join, exists, split
+import copy
+import logging
+import xmlschema
+
+from .report import Report
+from .record import PROBLEM, Record, INFORMATION
+
+XML_SCHEMA = join(dirname(__file__), 'xml_schema')
+LOGGER = logging.getLogger(__name__)
+
+
+def schemas(report: Report, CONST_parsed_xml, branch_name):
+    """validates XML file with existing schemas
+    :parsed_xml: parsed data from an XML file
+    """
+    parsed_xml = copy.copy(CONST_parsed_xml)
+    failed, metadatacount, valid = _validation_checks(report, parsed_xml, branch_name)
+
+    root_schema = join(XML_SCHEMA, 'addon.xsd')
+    if not _validate(parsed_xml, root_schema):
+        report.add(Record(PROBLEM, "Schema validation failed for root addon element"))
+        valid = False
+
+    if metadatacount == 0:
+        report.add(Record(PROBLEM, 'Metadata missing/occurred more than once'))
+
+    if valid:
+        report.add(Record(INFORMATION, "Valid XML file found"))
+    elif failed:
+        report.add(Record(PROBLEM, "Schema validation failed for the following points: %s " % ''.join(failed)))
+
+
+def _validation_checks(report: Report, parsed_xml, branch_name):
+    """Perform certain checks on the XML elements and extension point
+    """
+    metadatacount = 0
+    valid = True
+    failed = []
+    valid_points = {
+        'kodi.context.item': 'contextitem.xsd',
+        'kodi.game.controller': 'controller.xsd',
+        'kodi.resource.games': 'games.xsd',
+        'kodi.resource.images': 'images.xsd',
+        'kodi.resource.language': 'language.xsd',
+        'kodi.addon.metadata': 'metadata.xsd',
+        'kodi.resource.uisounds': 'uisounds.xsd',
+        'xbmc.addon.metadata': 'metadata.xsd',
+        'xbmc.addon.repository': 'repository.xsd',
+        'xbmc.gui.skin': 'skin.xsd',
+        'xbmc.metadata.scraper.albums': 'scraper.xsd',
+        'xbmc.metadata.scraper.artists': 'scraper.xsd',
+        'xbmc.metadata.scraper.movies': 'scraper.xsd',
+        'xbmc.metadata.scraper.musicvideos': 'scraper.xsd',
+        'xbmc.metadata.scraper.tvshows': 'scraper.xsd',
+        'xbmc.metadata.scraper.library': 'scraper.xsd',
+        'xbmc.python.script': 'script.xsd',
+        'xbmc.python.lyrics': 'script.xsd',
+        'xbmc.python.weather': 'script.xsd',
+        'xbmc.python.library': 'script.xsd',
+        'xbmc.python.pluginsource': 'pluginsource.xsd',
+        'xbmc.python.module': 'script.xsd',
+        'xbmc.service': 'service.xsd',
+        'xbmc.subtitle.module': 'script.xsd',
+        'xbmc.ui.screensaver': 'script.xsd',
+        'xbmc.webinterface': 'webinterface.xsd',
+    }
+
+    for extension in parsed_xml.findall("extension"):
+        extension_point = extension.get("point")
+        if extension_point in valid_points:
+            schema_path = check_version(branch_name, valid_points[extension_point])
+
+            if schema_path:
+                if extension_point == "xbmc.addon.metadata" or extension_point == "kodi.addon.metadata":
+                    metadatacount += 1
+
+                if not exists(schema_path):
+                    report.add(Record(PROBLEM, "%s doesn't exist" % schema_path))
+                    continue
+
+                if not _validate(extension, schema_path):
+                    failed.append(extension_point)
+                    valid = False
+
+                parsed_xml.remove(extension)
+            else:
+                report.add(Record(PROBLEM, "schema for %s doesn't exists" % extension_point))
+        else:
+            report.add(Record(PROBLEM, "%s is not a valid extension point" % extension_point))
+            valid = False
+
+    return failed, metadatacount, valid
+
+
+def _validate(xml, schemapath):
+    """Validates given XML file/element with a given schema
+
+    :xml: parsed xml element/file
+    :schemapath: path to a schema file
+    """
+    schema = xmlschema.XMLSchema(schemapath)
+    for error in schema.iter_errors(xml):
+        print(error)
+    return schema.is_valid(xml)
+
+
+def check_version(branch_name, schema_file):
+    all_branches = ['leia', 'krypton', 'jarvis', 'isengard', 'helix', 'gotham']
+    branches = all_branches[all_branches.index(branch_name)::1]
+    for branch in branches:
+        file = branch + '_' + schema_file
+        schema_file_path = join(XML_SCHEMA, file)
+        if exists(schema_file_path):
+            return schema_file_path
+    return None
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ It can also be used locally for detecting problems in your addons.
 
 - Check specific [version attribute](https://kodi.wiki/view/Addon.xml#version_attribute_2)
 
+- Validate addon.xml against schemas
+
 All of the validation and checks are done according to the kodi [addon rules](https://kodi.wiki/view/Add-on_rules)
 
 ## Installation
--- a/kodi_addon_checker/xml_schema/addon.xsd
+++ b/kodi_addon_checker/xml_schema/addon.xsd
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:element name="addon">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element name="requires" type="requiresType" minOccurs="0" maxOccurs="1"/>
+      </xs:sequence>
+      <xs:attribute name="id" type="xs:string" use="required"/>
+      <xs:attribute name="name" type="xs:string" use="required"/>
+      <xs:attribute name="version" type="xs:string" use="required"/>
+      <xs:attribute name="provider-name" type="xs:string" use="required"/>
+    </xs:complexType>
+  </xs:element>
+  <xs:complexType name="requiresType">
+    <xs:sequence>
+      <xs:element name="import" type="importType" minOccurs="0" maxOccurs="unbounded"/>
+    </xs:sequence>
+  </xs:complexType>
+  <xs:complexType name="importType">
+    <xs:attribute name="addon" type="xs:string" use="required"/>
+    <xs:attribute name="version" type="xs:string"/>
+    <xs:attribute name="optional" type="xs:boolean"/>
+  </xs:complexType>
+</xs:schema>
--- a/kodi_addon_checker/xml_schema/gotham_images.xsd
+++ b/kodi_addon_checker/xml_schema/gotham_images.xsd
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE schema PUBLIC "-//W3C//DTD XMLSCHEMA 200102//EN" "http://www.w3.org/2001/XMLSchema.dtd">
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:element name="extension">
+    <xs:complexType>
+      <xs:attribute name="point" type="simpleIdentifier" use="required"/>
+      <xs:attribute name="compile" type="xs:boolean"/>
+      <xs:attribute name="type" type="xs:string"/>
+    </xs:complexType>
+  </xs:element>
+  <xs:simpleType name="simpleIdentifier">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="kodi\.resource\.images"/>
+    </xs:restriction>
+  </xs:simpleType>
+</xs:schema>
--- a/kodi_addon_checker/xml_schema/gotham_metadata.xsd
+++ b/kodi_addon_checker/xml_schema/gotham_metadata.xsd
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE schema PUBLIC "-//W3C//DTD XMLSCHEMA 200102//EN" "http://www.w3.org/2001/XMLSchema.dtd">
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:element name="extension">
+    <xs:complexType>
+      <xs:choice maxOccurs="unbounded">
+        <xs:element name="summary" type="translatedString" minOccurs="1" maxOccurs="unbounded"/>
+        <xs:element name="description" type="translatedString" minOccurs="0" maxOccurs="unbounded"/>
+        <xs:element name="disclaimer" type="translatedString" minOccurs="0" maxOccurs="unbounded"/>
+        <xs:element name="platform" type="platformList" minOccurs="0"/>
+        <xs:element name="language" type="xs:string"  minOccurs="0" maxOccurs="1"/>
+        <xs:element name="license" type="xs:string"  minOccurs="0" maxOccurs="1" />
+        <xs:element name="forum" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="website" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="source" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="email" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="broken" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="news" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="reuselanguageinvoker" type="xs:boolean" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="assets" type="assetsList" minOccurs="0" maxOccurs="1"/>
+      </xs:choice>
+      <xs:attribute name="point" type="xs:string" use="required"/>
+      <xs:attribute name="id" type="simpleIdentifier"/>
+      <xs:attribute name="name" type="xs:string"/>
+    </xs:complexType>
+  </xs:element>
+  <xs:simpleType name="simpleIdentifier">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="(kodi|xbmc)\.addon\.metadata"/>
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:complexType name="translatedString">
+    <xs:simpleContent>
+      <xs:extension base="xs:string">
+        <xs:attribute name="lang" type="langIdentifier" use="required"/>
+      </xs:extension>
+    </xs:simpleContent>
+  </xs:complexType>
+  <xs:simpleType name="platformType">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="linux"/>
+      <xs:enumeration value="osx"/>
+      <xs:enumeration value="osx64"/>
+      <xs:enumeration value="osx32"/>
+      <xs:enumeration value="ios"/>
+      <xs:enumeration value="windx"/>
+      <xs:enumeration value="windows"/>
+      <xs:enumeration value="windowsstore"/>
+      <xs:enumeration value="android"/>
+      <xs:enumeration value="freebsd"/>
+      <xs:enumeration value="all"/>
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="langIdentifier">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="[a-z]{2,3}(_[A-Z]{2}(@\S+)?)?"/>
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:complexType name="assetsList">
+    <xs:choice maxOccurs="unbounded">
+        <xs:element name="icon" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="fanart" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="screenshot" type="xs:string" minOccurs="0"/>
+        <xs:element name="clearlogo" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="banner" type="xs:string" minOccurs="0" maxOccurs="1"/>
+    </xs:choice>
+  </xs:complexType>
+  <xs:simpleType name="platformList">
+    <xs:list itemType="platformType"/>
+  </xs:simpleType>
+</xs:schema>
--- a/kodi_addon_checker/xml_schema/gotham_pluginsource.xsd
+++ b/kodi_addon_checker/xml_schema/gotham_pluginsource.xsd
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE schema PUBLIC "-//W3C//DTD XMLSCHEMA 200102//EN" "http://www.w3.org/2001/XMLSchema.dtd">
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:element name="extension">
+    <xs:complexType>
+      <xs:choice maxOccurs="unbounded">
+        <xs:element name="provides" type="providesList" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="medialibraryscanpath" type="pathType" minOccurs="0" maxOccurs="unbounded"/>
+      </xs:choice>
+      <xs:attribute name="point" type="xs:string" use="required"/>
+      <xs:attribute name="id" type="simpleIdentifier"/>
+      <xs:attribute name="name" type="xs:string"/>
+      <xs:attribute name="library" type="xs:string" use="required"/>
+    </xs:complexType>
+  </xs:element>
+  <xs:simpleType name="simpleIdentifier">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="xbmc\.python\.pluginsource"/>
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="providesType">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="audio"/>
+      <xs:enumeration value="executable"/>
+      <xs:enumeration value="game"/>
+      <xs:enumeration value="image"/>
+      <xs:enumeration value="video"/>
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="providesList">
+    <xs:list itemType="providesType"/>
+  </xs:simpleType>
+  <xs:simpleType name="contentType">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="albums"/>
+      <xs:enumeration value="artists"/>
+      <xs:enumeration value="movies"/>
+      <xs:enumeration value="musicvideos"/>
+      <xs:enumeration value="tvshows"/>
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:complexType name="pathType">
+    <xs:simpleContent>
+      <xs:extension base="xs:string">
+        <xs:attribute name="content" type="contentType" use="required"/>
+      </xs:extension>
+    </xs:simpleContent>
+  </xs:complexType>
+</xs:schema>
--- a/kodi_addon_checker/xml_schema/gotham_repository.xsd
+++ b/kodi_addon_checker/xml_schema/gotham_repository.xsd
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE schema PUBLIC "-//W3C//DTD XMLSCHEMA 200102//EN" "http://www.w3.org/2001/XMLSchema.dtd">
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:element name="extension">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element name="dir" type="dirItem" minOccurs="1" maxOccurs="unbounded"/>
+      </xs:sequence>
+      <xs:attribute name="point" type="xs:string" use="required"/>
+      <xs:attribute name="id" type="simpleIdentifier"/>
+      <xs:attribute name="name" type="xs:string"/>
+    </xs:complexType>
+  </xs:element>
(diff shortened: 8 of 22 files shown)


MartijnKaijser:
i'm against pulling in these schemes cause if they change we would need to sync them again. Best would be to pull them from xbmc git on the fly

mzfr:
These xsd come from different folders in the xbmc repo. Do you want me to put the link to these in the code?

tamland:
The schemas aren't complete though. They're only for the 'extension-point' elements, and not even all of them. Maybe you can fix this in the main repo and make it complete? 

Feel free to move them all into one file too. (But imo is not really a big deal to copy-paste. I changes like once a year)

Razzeee:
@tamland do we use the schemas for anything at the moment? Just wondering as you point out that they are not complete

tamland:
@Razzeee no, they're just used as documentation really.

mzfr:
@tamland What do you mean by schemas are not complete ?And what all schemas are missing ?

tamland:
I mean the schemas are just what's inside the extention-point elements. There's no schema for the `<addon>` element. And the metadata one is missing `assets` for instance. (you'll have to check against source for the complete list)

mzfr:
@tamland I am willing to complete the schemas but I am really not sure how I am supposed to do it. So maybe you can provide some kind of outline for them or maybe you can complete any one of the schemas and then I'll use it as a reference for completing other schemas.

tamland:
@mzfr You need to write the actual schema definition so they can be used to verify addon.xml (thats' what you want to achieve right?) Here's a start that checks `<addon>`, `<requires>` and `<import>`:
http://www.utilities-online.info/xsdvalidation/?save=bc408307-8a02-4aab-908a-ae4141c02bdc-xsdvalidation
Do the same for `<extension>` element: defined types for it and include everything from the existing .xsd files to make it complete.

tamland:
Here's another example the verifies that `<extension point="xbmc.addon.metadata">` occurs exactly once: http://www.utilities-online.info/xsdvalidation/?save=60fc9516-796d-4ccb-b610-9d570261d4a2-xsdvalidation (probably a million different ways you can do this; this is just *one* example)

mzfr:
@tamland yes that's what we are trying to achieve. Thanks a lot for the examples they will be very helpful :slightly_smiling_face:

kodi_addon_checker/check_addon.py
```diff
@@ -124,3 +127,15 @@ def _get_addons(xml_url: str):
         LOGGER.error(errrt)
     except requests.exceptions.ConnectTimeout as errct:
         LOGGER.error(errct)
+
+
+def check_schema(report: Report, Addon_xml, schema_path):
+    for xsd in os.listdir(schema_path):
+        schema = xmlschema.XMLSchema(schema_path + "/" + xsd)
+        try:
+            if schema.validate(Addon_xml):
```
Rechi: validate doesn't return anything, it only raises exceptions

kodi_addon_checker/check_addon.py
```diff
@@ -124,3 +127,15 @@ def _get_addons(xml_url: str):
         LOGGER.error(errrt)
     except requests.exceptions.ConnectTimeout as errct:
         LOGGER.error(errct)
+
+
+def check_schema(report: Report, Addon_xml, schema_path):
+    for xsd in os.listdir(schema_path):
```
Rechi: why?

kodi_addon_checker/check_addon.py
```diff
@@ -124,3 +127,15 @@ def _get_addons(xml_url: str):
...
+        schema = xmlschema.XMLSchema(schema_path + "/" + xsd)
+        try:
+            if schema.validate(Addon_xml):
+                report.add(Record(INFORMATION, "Valid XML file found"))
+
+        except xmlschema.XMLSchemaValidationError as schemaerror:
+            print(schemaerror)
+            report.add(Record(INFORMATION, "Schema validation failed"))
```
Rechi: ERROR or at least WARNING

kodi_addon_checker/check_addon.py
mzfr: First I was thinking that we'll test a single addon.xml against all the schema that is why I added this loop.
But now I realized that we don't need it instead we will have to test every addon.xml according to their extension point right ?

kodi_addon_checker/check_addon.py
mzfr: I have added validate just for now. I'll replace it with `schema.is_valid` after everything works

kodi_addon_checker/check_addon.py
Rechi: Can't you write one xsd schema which is correct for all addons?

kodi_addon_checker/check_addon.py
Rechi: But then you can't tell what's wrong.

kodi_addon_checker/check_addon.py
mzfr: for all addons as in for all the extension point i.e `plugins`, `scripts`, `resources` ?

Rechi: Yes on schema which contains everything for plugins, script, resources, skins (and maybe more, don't know if we have more).

mzfr: okay I'll try to do so

kodi_addon_checker/check_addon.py
```diff
@@ -124,3 +127,16 @@ def _get_addons(xml_url: str):
...
     except requests.exceptions.ConnectTimeout as errct:
         LOGGER.error(errct)
+
+
+def check_schema(report: Report, Addon_xml, schema_path):
+    schema = xmlschema.XMLSchema(schema_path)
+
+    try:
+        schema.validate(Addon_xml)
+        if schema.is_valid(Addon_xml):
```
Rechi: This if is is useless, either validate throws and exception or not. If it doesn't the xml is valid.

mzfr: So basically we don't want any output when the schema is valid ?

Rechi: I didn't say that, just saying if you hit this line you already know it is valid, because otherwise it you jump into the except block.

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,187 @@
...
+  <xs:element name="addon">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element ref="requires"/>
+        <xs:element maxOccurs="unbounded" ref="extension"/>
+        <xs:sequence minOccurs="0">
+          <xs:element ref="news"/>
+          <xs:element ref="license"/>
+          <xs:element ref="source"/>
+        </xs:sequence>
```
Rechi: elements not allowed inside the addon tag (only inside element tag)

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,181 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified">
+  <xs:element name="addon">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element ref="requires" minOccurs="0"/>
+        <xs:element maxOccurs="unbounded" ref="extension"/>
+        <xs:element ref="news"/>
```
Rechi: not allowed here

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,181 @@
...
+        <xs:element ref="license"/>
```
Rechi: not allowed here

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,181 @@
...
+        <xs:element ref="source"/>
```
Rechi: not allowed here

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,181 @@
...
+        <xs:element minOccurs="0" ref="assets"/>
```
Rechi: not allowed here

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,181 @@
...
+          <xs:element ref="email"/>
+          <xs:element ref="forum"/>
+          <xs:element ref="language"/>
+          <xs:element ref="platform"/>
+          <xs:element ref="summary"/>
+          <xs:element ref="website"/>
+        </xs:choice>
+        <xs:choice minOccurs="0">
+          <xs:element ref="licence"/>
+          <xs:element ref="nofanart"/>
```
Rechi: only repo generator adds this

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,181 @@
...
+      <xs:attribute name="library" use="required" type="xs:NCName"/>
+    </xs:complexType>
+  </xs:element>
+  <xs:element name="label" type="xs:integer"/>
+  <xs:element name="visible" type="xs:string"/>
+  <xs:element name="source" type="xs:anyURI"/>
+  <xs:element name="license" type="xs:string"/>
+  <xs:element name="assets">
+    <xs:complexType>
+      <xs:sequence>
```
Rechi: no specific order for assets

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,168 @@
...
+  <xs:element name="summary">
+    <xs:complexType mixed="true">
+      <xs:attribute name="lang" use="required" type="xs:NCName"/>
+    </xs:complexType>
+  </xs:element>
+  <xs:element name="news" type="xs:string"/>
+  <xs:element name="website" type="xs:anyURI"/>
+  <xs:element name="licence" type="xs:string"/>
+  <xs:element name="nofanart" type="xs:boolean"/>
+  <xs:element name="provides" type="xs:string"/>
```
enen92: This should be restricted to the allowed values, see: https://github.com/xbmc/xbmc/blob/8d4a5bba55638dfd0bdc5e7de34f3e5293f99933/addons/xbmc.python/pluginsource.xsd#L20

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,168 @@
...
+          <xs:element ref="forum"/>
+          <xs:element ref="language"/>
+          <xs:element ref="platform"/>
+          <xs:element ref="summary"/>
+          <xs:element ref="website"/>
+        </xs:choice>
+      </xs:sequence>
+      <xs:attribute name="compile" type="xs:boolean"/>
+      <xs:attribute name="defaultresolution" type="xs:NMTOKEN"/>
+      <xs:attribute name="library"/>
```
enen92: library is a required attribute

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,168 @@
...
+  <xs:element name="platform" type="xs:string"/>
+  <xs:element name="summary">
+    <xs:complexType mixed="true">
+      <xs:attribute name="lang" use="required" type="xs:NCName"/>
+    </xs:complexType>
+  </xs:element>
+  <xs:element name="news" type="xs:string"/>
+  <xs:element name="website" type="xs:anyURI"/>
+  <xs:element name="licence" type="xs:string"/>
+  <xs:element name="nofanart" type="xs:boolean"/>
```
enen92: According to AddonManager you're missing `noicon` too: https://github.com/xbmc/xbmc/blob/0b8f23abe8b9e637071107f3c0c8dac8e89b7b9a/xbmc/addons/AddonManager.cpp#L189

xml_schema/plugin_addon.xsd
```diff
@@ -0,0 +1,168 @@
...
+          <xs:element ref="email"/>
+          <xs:element ref="forum"/>
+          <xs:element ref="language"/>
+          <xs:element ref="platform"/>
+          <xs:element ref="summary"/>
+          <xs:element ref="website"/>
+        </xs:choice>
+      </xs:sequence>
+      <xs:attribute name="compile" type="xs:boolean"/>
+      <xs:attribute name="defaultresolution" type="xs:NMTOKEN"/>
```
enen92: this is only used in skins not in plugins

xml_schema/plugin_addon.xsd
Rechi: no, only repo generator adds nofanart and noicon

enen92: so `nofanart` should go away :)

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,33 @@
...
+from .record import PROBLEM, Record, INFORMATION
+
+xml_schema = "./xml_schema/"
+LOGGER = logging.getLogger(__name__)
+
+
+def schemas(report: Report, parsed_xml):
+
+    for extension in parsed_xml.findall("extension"):
+        extension_point = (extension.get("point")).split(".")[-1]
```
enen92: you can't expect all provided schemas will be valid. The point attribute might not exist, this will crash your tool

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,33 @@
...
+        schema_path = xml_schema + extension_point + '.xsd'
```
enen92: use `os.path.join` instead of concatenating strings manually

kodi_addon_checker/schema_validation.py
mzfr: then we can either have a list to check if the point exists or not. Or we can use `try` and `expect`.

enen92: yes and also other things:
- All addons MUST have a metadata extension element
- All addons MUST at least define one of the other extension elements

If the above conditions do not hold (plus the possibility of an incorrect `extension` element without the point attribute) you can fail the overall validation right away

Rechi: You mean add some early returns to this function? I disagree with that. The checker should always report all issues and not only some till you fix one.

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,33 @@
...
+    validate(report, parsed_xml, root_schema)
+
+
+def validate(report: Report, xml, schemapath):
+
+    schema = xmlschema.XMLSchema(schemapath)
+
+    try:
+        schema.validate(xml)
+        report.add(Record(INFORMATION, "Valid XML file found"))
```
Rechi: useless info, you expect that if no issue is logged it is valid

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,33 @@
...
+
+    except xmlschema.XMLSchemaValidationError as schemaerror:
+        LOGGER.error(schemaerror)
+        report.add(Record(PROBLEM, "Schema validation failed"))
```
Rechi: report the reason, otherwise you have to guess what's wrong

kodi_addon_checker/schema_validation.py
enen92: The comment was more directed towards introducing a few more validation steps to mzfr than deciding upon that. I agree the checker should report all problems even if an early one is found.

kodi_addon_checker/schema_validation.py
Rechi: If you want to report that it is valid then this is the wrong place as this might be called multiple times for the same addon.xml. It could also happen that it reports all fine and an error for the same addon which would be confusing.

kodi_addon_checker/schema_validation.py
Rechi: Log file doesn't help as you won't see that on travis.

mzfr: I did that to keep the travis log clean. If you want I'll add the report the reason to normal logs.

xml_schema/addon.xsd
```diff
@@ -0,0 +1,24 @@
...
+    </xs:complexType>
+  </xs:element>
+  <xs:complexType name="requiresType">
+    <xs:sequence>
+      <xs:element name="import" type="importType" minOccurs="0" maxOccurs="unbounded"/>
+    </xs:sequence>
+  </xs:complexType>
+  <xs:complexType name="importType">
+    <xs:attribute name="addon" type="xs:string" use="required"/>
+    <xs:attribute name="version" type="xs:string"/>
```
enen92: please add:
`<xs:attribute name="optional" type="xs:boolean"/>`

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,47 @@
...
+
+xml_schema = "./xml_schema/"
+LOGGER = logging.getLogger(__name__)
+
+
+def schemas(report: Report, parsed_xml):
+    metadatacount = 0
+    validcount = 1
+    for extension in parsed_xml.findall("extension"):
+        extension_point = (extension.get("point")).split(".")[-1]
```
enen92: this fails at least for context menu extension points. Example:

´´´xml
  <extension point="kodi.context.item">
    <menu id="kodi.core.manage">
      <item library="addon.py">
        <label>32000</label>
        <visible>String.IsEqual(ListItem.dbtype,movie) | String.IsEqual(ListItem.dbtype,tvshow)</visible>
      </item>
    </menu>
  </extension>
```

File: kodi_addon_checker/schema_validation.py [hunk 22]
```diff
@@ -0,0 +1,47 @@
...
+        if extension_point == "metadata":
+            metadatacount += 1
+
+        schema_path = xml_schema + extension_point + '.xsd'
```
enen92: I still don't like this use `os.path.join` to build the final path for the xsd file. Does `xml_schema = "./xml_schema/"` even work on windows? Obtain the relative path to the schema folder using the `os` module too

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,47 @@
...
+        if not os.path.exists(schema_path):
+            report.add(Record(PROBLEM, "%s doesn't exist" % extension_point))
+            continue
+
+        if not validate(report, extension, schema_path):
+            report.add(Record(PROBLEM, "Schema validation failed for %s" % extension_point))
+            validcount -= 1
+        parsed_xml.remove(extension)
+
+    root_schema = xml_schema + 'addon.xsd'
```
enen92: same here

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,47 @@
...
+        report.add(Record(INFORMATION, "Valid XML file found"))
+
+
+def validate(report: Report, xml, schemapath):
+    try:
+        schema = xmlschema.XMLSchema(schemapath)
+        schema.validate(xml)
+        return True
+    except xmlschema.XMLSchemaValidationError as schemaerror:
+        print(schemaerror)
```
enen92: You're passing a report as the argument, why do you do a print? What if we want to use another reporter in the future?

kodi_addon_checker/schema_validation.py
mzfr: is it failing only for contextitem only ? If yes then I can add another if statement.
Something like
```python
if extension_point == "item":
   extension_point = "contextitem"
```

kodi_addon_checker/schema_validation.py
mzfr: I forgot to remove report from the argument. I have used print because if we add this in our report as a `PROBLEM` that will increase the count of errors (`ERROR: We found 1 problems and 0 warnings, please check the logfile.`) twice for a same problem.

kodi_addon_checker/schema_validation.py
enen92: There are also other problems you should take into account, this is not as easy as a new if statement. You need to create some sort of mapping between extension points and schemas. From kodi source tree:
```
  <extension-point id="script" schema="script.xsd"/>
  <extension-point id="subtitles" schema="script.xsd"/>
  <extension-point id="lyrics" schema="script.xsd"/>
  <extension-point id="weather" schema="script.xsd"/>
  <extension-point id="library" schema="script.xsd"/>
  <extension-point id="screensaver" schema="script.xsd"/>
  <extension-point id="plugin" schema="pluginsource.xsd"/>
  <extension-point id="context.item" schema="contextitem.xsd"/>
```

mzfr: After looking at this I think having a dictionary lookup would be nice

kodi_addon_checker/schema_validation.py
enen92: Agreed. Please remove the reporter from the method arguments if not used

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,47 @@
...
+            report.add(Record(PROBLEM, "%s doesn't exist" % extension_point))
+            continue
+
+        if not validate(report, extension, schema_path):
+            report.add(Record(PROBLEM, "Schema validation failed for %s" % extension_point))
+            validcount -= 1
+        parsed_xml.remove(extension)
+
+    root_schema = xml_schema + 'addon.xsd'
+    validate(report, parsed_xml, root_schema)
```
enen92: you're not decreasing the validcount var if root_schema validation fails

tamland:
Why are we manually splitting out the `extension` elements and validating those individually? Why not make this part of addon.xsd and validate the whole thing in one go?

enen92:
Yes ideally we wanted to have a single schema but also to reuse the ones we already have (complete them if necessary) since they are part of Kodi. `Include` revealed to be a problem as everything is in the same namespace and referring to elements is only possible with the `ref` attribute (which performs the lookup by name). So this was basically a decision to reduce the maintainability of the schemas and to have something working by the time the GSOC project ends.

Hopefully, at some point in the future and with xsd 1.1 we can have a better approach. One that can rely on the `test` attribute to check for different extension schemas depending on the extension `point` and requiring minimal changes to each of the extension schemas we already have. The problem is that there are no python tools at the moment that are able to validate xsd 1.1 (the `xmlschema` module we are using have in on the roadmap though).

What we are doing to validate the schemas is not ideally but it was a pondered decision. If you have a better idea/suggestion let us know @tamland.

xml_schema/addon.xsd
enen92: Do not forget this one ^^. This is your schema, the only one that is not in kodi's github.

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def _validation_checks(report: Report, parsed_xml):
+    """Perform certain checks on the XML elements and extension point
+    """
+    metadatacount = 0
+    validcount = 1
+    failed = []
+    variation = ['kodi', 'xbmc']
+    valid_points = {'metadata': 'metadata.xsd', 'pluginsource': 'pluginsource.xsd', 'plugin': 'pluginsource.xsd',
```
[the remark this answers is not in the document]
mzfr: So you want it like `{'xbmc.python.script': 'script.xsd'} ?` also should I add them all ?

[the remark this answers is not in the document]
enen92: You should add the complete `point` attribute values for the schemas you have. Basically anything that can be submitted to `repo-plugins` (plugins) and `repo-scripts` (services, context.item, scripts, python.modules, subtitles, weather, lyrics ) is top priority (don't forget the metadata). Others (`repo-resources`, `skins`) are nice to have but not as important as the first. Try to map what you have right now to the complete names on the list above.

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,89 @@
...
+                    'xbmc.metadata.scraper.movies': 'scraper.xsd', 'xbmc.metadata.scraper.musicvideos': 'scraper.xsd',
+                    'xbmc.metadata.scraper.tvshows': 'scraper.xsd', 'xbmc.metadata.scraper.library': 'scraper.xsd'}
+
+    for extension in parsed_xml.findall("extension"):
+        extension_point = (extension.get("point"))
+        point = extension_point.split(".")[-1]
+
+        if extension_point in valid_points:
+            schema_file = valid_points[extension_point]
+            if point == "metadata":
```
enen92: No need to do the split, check `xbmc.addon.metadata` directly

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,89 @@
...
+                metadatacount += 1
+
+            schema_path = os.path.join(XML_SCHEMA, schema_file)
```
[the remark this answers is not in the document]
mzfr: if we want something(report/logging) to happen if the file is missing then why not just leave it the way it is ?

xml_schema/addon.xsd
mzfr: Yes will do it the moment PR #14237 merge

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,88 @@
...
+        report.add(Record(PROBLEM, "Schema validation failed for the following points: %s " % ''.join(failed)))
+
+
+def _validation_checks(report: Report, parsed_xml):
+    """Perform certain checks on the XML elements and extension point
+    """
+    metadatacount = 0
+    validcount = 1
+    failed = []
+    valid_points = {'xbmc.addon.metadata': 'metadata.xsd', 'xbmc.python.pluginsource': 'pluginsource.xsd',
```
enen92: @mzfr `point="xbmc.python.module"` missing -> should point to script.xsd too

enen92: `kodi.addon.metadata` also missing (same as xbmc.addon.metadata)

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,88 @@
...
+                    'xbmc.metadata.scraper.albums': 'scraper.xsd', 'xbmc.metadata.scraper.artists': 'scraper.xsd',
+                    'xbmc.metadata.scraper.movies': 'scraper.xsd', 'xbmc.metadata.scraper.musicvideos': 'scraper.xsd',
+                    'xbmc.metadata.scraper.tvshows': 'scraper.xsd', 'xbmc.metadata.scraper.library': 'scraper.xsd'}
+
+    for extension in parsed_xml.findall("extension"):
+        extension_point = extension.get("point")
+
+        if extension_point in valid_points:
+            schema_file = valid_points[extension_point]
+            if extension_point == "xbmc.addon.metadata":
```
enen92: check also `kodi.addon.metadata`

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,88 @@
...
+    if not _validate(parsed_xml, root_schema):
+        report.add(Record(PROBLEM, "Schema validation failed for root addon element"))
+        validcount -= 1
+
+    if metadatacount != 1:
+        report.add(Record(PROBLEM, 'Metadata missing/occurred more than once'))
+
+    if validcount == 1:
+        report.add(Record(INFORMATION, "Valid XML file found"))
+    else:
```
enen92: log only if failed is not empty

xml_schema/controller.xsd
```diff
@@ -0,0 +1,23 @@
+
```
enen92: the empty line is breaking controller addons schema validation (and the error is not upstream). Please make sure you also fix it in next iteration.

mzfr: Next iteration meaning ?

enen92: I identified a few issues with current schemas in our tests (see logs in https://travis-ci.org/xbmc/addon-check/builds/409912337 and identified issues in https://github.com/xbmc/xbmc/pull/14237#issuecomment-409546715). When those are fixed we'll try again to see if everything is okay.

kodi_addon_checker/schema_validation.py
mzfr: DO you mean to say I should remove `validcount` ?

enen92: No, I mean instead of the else do a `elif failed:` so you only report if actually have a failed list.

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,100 @@
...
+        report.add(Record(PROBLEM, "Schema validation failed for the following points: %s " % ''.join(failed)))
+
+
+def _validation_checks(report: Report, parsed_xml, branch_name):
+    """Perform certain checks on the XML elements and extension point
+    """
+    metadatacount = 0
+    validcount = 1
+    failed = []
+    valid_points = {'xbmc.python.module': 'script.xsd', 'kodi.addon.metadata': 'metadata.xsd',
```
enen92: please keep one `key:value` per line to improve readability. Some keys are repeated (kodi.addon.metadata and xbmc.python.module)

Rechi: also sort them somehow, best would be sort by value and then per key

xml_schema/isengard_contextitem.xsd
```diff
@@ -0,0 +1,30 @@
...
+  <xs:complexType name="itemType">
+    <xs:all>
+      <xs:element name="label" type="xs:string" minOccurs="1"/>
+      <xs:element name="visible" type="xs:string" minOccurs="1"/>
+      <xs:element name="parent" type="parentIdentifier" minOccurs="0"/>
+    </xs:all>
+  </xs:complexType>
+  <xs:simpleType name="pointIdentifier">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="kodi.context.item"/>
```
enen92: `kodi\.context\.item` for consistency with the other schemas (after rechi's comments)

xml_schema/isengard_contextitem.xsd
```diff
@@ -0,0 +1,30 @@
...
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="parentIdentifier">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="kodi.core.main|kodi.core.manage"/>
```
enen92: `kodi\.core\.(manage|main)`

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,114 @@
...
+    :xml: parsed xml element/file
+    :schemapath: path to a schema file
+    """
+    try:
+        schema = xmlschema.XMLSchema(schemapath)
+        schema.validate(xml)
+        return True
+    except xmlschema.XMLSchemaValidationError as schemaerror:
+        print(schemaerror)
+        return False
```
Rechi: replace the whole try/catch logic with the following to get all validation errors and not only the first one
```
schema = xmlschema.XMLSchema(schemapath)
for error in schema.iter_errors(xml):
    print(error)
return schema.is_valid(xml)
```

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,114 @@
...
+        'xbmc.python.module': 'script.xsd',
+        'xbmc.service': 'service.xsd',
+        'xbmc.subtitle.module': 'script.xsd',
+        'xbmc.ui.screensaver': 'script.xsd',
+        'xbmc.webinterface': 'webinterface.xsd',
+    }
+
+    for extension in parsed_xml.findall("extension"):
+        extension_point = extension.get("point")
+        schema_path = check_version(branch_name, valid_points[extension_point])
```
Rechi: `extension_point in valid_points` has to be checked before this line

xml_schema/language.xsd
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
enen92: which schema is this? IIRC we only need isengard_language

Rechi:
schema validation for addon.xml has to be added to README.md

mzfr:
@enen92 I have squashed all the commits

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,115 @@
...
+                    metadatacount += 1
+
+                if not os.path.exists(schema_path):
+                    report.add(Record(PROBLEM, "%s doesn't exist" % schema_path))
+                    continue
+
+                if not _validate(extension, schema_path):
+                    failed.append(extension_point)
+                    validcount -= 1
+                parsed_xml.remove(extension)
```
Rechi: This changes `parsed_xml` passed at check_addon.py and causes an incomplete view of the addon.xml for later checks.

Rechi: While testing #108 onto this PR I wondered why no assets are available for the transparency checks. Following change proves my observation.
```
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -43,5 +43,7 @@
             file_index = handle_files.create_file_index(addon_path)

+            print(ET.tostring(parsed_xml))
             schema_validation.schemas(addon_report, parsed_xml, branch_name)
+            print(ET.tostring(parsed_xml))

             check_dependencies.check_addon_dependencies(addon_report, repo_addons, parsed_xml, branch_name)
```

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,115 @@
...
+        report.add(Record(INFORMATION, "Valid XML file found"))
+    elif failed:
+        report.add(Record(PROBLEM, "Schema validation failed for the following points: %s " % ''.join(failed)))
+
+
+def _validation_checks(report: Report, parsed_xml, branch_name):
+    """Perform certain checks on the XML elements and extension point
+    """
+    metadatacount = 0
+    validcount = 1
```
Rechi: this could be a boolean

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,115 @@
...
+XML_SCHEMA = join(HERE, 'xml_schema')
+LOGGER = logging.getLogger(__name__)
+
+
+def schemas(report: Report, CONST_parsed_xml, branch_name):
+    """validates XML file with existing schemas
+    :parsed_xml: parsed data from an XML file
+    """
+    parsed_xml = copy.copy(CONST_parsed_xml)
+    failed, metadatacount, validcount = _validation_checks(report, parsed_xml, branch_name)
```
Rechi: rename `validcount ` to `valid`, as this var doesn't count anything

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,115 @@
...
+        report.add(Record(INFORMATION, "Valid XML file found"))
+    elif failed:
+        report.add(Record(PROBLEM, "Schema validation failed for the following points: %s " % ''.join(failed)))
+
+
+def _validation_checks(report: Report, parsed_xml, branch_name):
+    """Perform certain checks on the XML elements and extension point
+    """
+    metadatacount = 0
+    validcount = True
```
Rechi: rename to `valid`, as this var doesn't count anything

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,115 @@
...
+        'xbmc.subtitle.module': 'script.xsd',
+        'xbmc.ui.screensaver': 'script.xsd',
+        'xbmc.webinterface': 'webinterface.xsd',
+    }
+    extensions = parsed_xml.findall("extension")
+    for extension in extensions:
+        extension_point = extension.get("point")
+        if extension_point in valid_points:
+            schema_path = check_version(branch_name, valid_points[extension_point])
+            print(schema_path)
```
Rechi: remove

kodi_addon_checker/schema_validation.py
```diff
@@ -0,0 +1,115 @@
...
+                if extension_point == "xbmc.addon.metadata" or extension_point == "kodi.addon.metadata":
+                    metadatacount += 1
+
+                if not exists(schema_path):
+                    report.add(Record(PROBLEM, "%s doesn't exist" % schema_path))
+                    continue
+
+                if not _validate(extension, schema_path):
+                    failed.append(extension_point)
+                    validcount = False
```
Rechi: `parsed_xml.remove(extension)` somehow got lost

PR #36, opened 2018-02-16
Check xml version for addon.xml

### LINKED ISSUE #6: Validate addon.xml beginning
Take any addon.xml. Change the header from this:

<?xml version="1.0" encoding="UTF-8" standalone="yes"?>

to these:

<?xml version="1.0.1" encoding="UTF-8" standalone="yes"?>

That used to break the repo update script.

mzfr:
Fixes #6 
I have done this by using lxml package.
I tried but couldn't figure out a way to find xml version using xml.etree.EelmentTree.

### DIFF
--- a/check_addon.py
+++ b/check_addon.py
@@ -2,6 +2,7 @@ import os
 import re
 import json
 import pathlib
+from lxml import etree
 import xml.etree.ElementTree
 from PIL import Image
 from common import colorPrint, check_config, has_transparency
@@ -61,6 +62,8 @@ def start(error_counter, addon_path, config = None):
             file_index = _create_file_index(addon_path)
 
             error_counter = _check_for_invalid_xml_files(error_counter, file_index)
+            error_counter = _check_addon_version(error_counter, addon_path)
+
 
             error_counter = _check_for_invalid_json_files(
                 error_counter, file_index)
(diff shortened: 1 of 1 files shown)


Rechi:
That's not the correct way to do it. Specification says `'1.[0-9]+` is valid (https://www.w3.org/TR/xml/#NT-VersionNum). Also we don't want hard-coded version number checks for it.

Have you checked why `repository-generator` fails if xml version is `1.0.1`?

mzfr:
@Rechi My understanding is that Python's `xml.etree` is a lenient parser as it parses this xml without raising any error:

```
<?xml version="1.0.1"?>
<addon>
    <name>Kodi</name>
</addon>
```

While lxml.etree raises a syntax error while parsing the same xml.

Since we use Python's `xml.etree` in addon-check and `lxml` in repository-generator, we pass `addon.xml` here which aren't really correct (acc. to lxml)

mzfr:
One way to deal with this is by using lxml here too, so that all such invalid XMLs just won't get into the system.
Or we could do something like what I've already suggested in this Pull.

To not hardcode a version here, we could use the Python's `re` module with `1.[0-9]+` regex pattern?

Rechi:
Just try to load `addon.xml` with `lxml`, no other checks are needed.

Rechi:
I just have noticed there is already a `_check_addon_xml` function. Can you move it into that function?
The current function name (`_check_addon_version`) is misleading as it doesn't check any version.

Razzeee:
Shouldn't this also fail for a schema check? https://github.com/xbmc/addon-check/pull/35
I'm confused that we need both, if that's the case. Haven't tried it, so just asking :)

Rechi:
Sure, schema checks should still fail, I was just speaking about the xml version check. There we shouldn't check for the version to match a regex.

Razzeee:
What I'm trying to say is that we can probably just not do this and solve
it by the schema check.

Razzeee:
@mzfr Can you agree that we close this one and focus on the schemas to validate the version?

mzfr:
@Razzeee  Yeah, that does seem like a better way to deal with this. Closing this!

PR #37, opened 2018-02-18
Strip Repository path from logs

### LINKED ISSUE #13: strip travis build paths from log
Maybe strip the travis paths from the log outputs like 
`
PROBLEM: Invalid xml found. /home/travis/build/xbmc/repo-plugins/plugin.video.reddit_viewer/resources/settings.xml`

mzfr:
Fixes #13 

Before this log output used to look like:
```
PROBLEM: Invalid xml found. /home/travis/build/xbmc/repo-plugins/
         plugin.video.reddit_viewer/resources/settings.xml
```

After applying these changes log output will look like:
```
PROBLEM: Invalid xml found. /xbmc/repo-plugins/
         plugin.video.reddit_viewer/resources/settings.xml
```

### DIFF
--- a/check_addon.py
+++ b/check_addon.py
@@ -6,7 +6,7 @@ import xml.etree.ElementTree
 from PIL import Image
 from common import colorPrint, check_config, has_transparency
 
-
+REL_PATH=""
 def _find_file(name, path):
     for file_name in os.listdir(path):
         match = re.match(name, file_name, re.IGNORECASE)
@@ -54,6 +54,10 @@ def start(error_counter, addon_path, config = None):
     colorPrint("Checking %s" % os.path.basename(
         os.path.normpath(addon_path)), "34")
 
+    global REL_PATH
+    # Extract common path from addon paths
+    # All paths will be printed relative to this path
+    REL_PATH = os.path.split(addon_path[:-1])[0]
     error_counter, addon_xml = _check_addon_xml(error_counter, addon_path)
 
     if addon_xml != None:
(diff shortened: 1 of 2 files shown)


Rechi:
IMO `/xbmc/repo-plugins/` should also be removed from output

mzfr:
@Rechi I have made the changes you suggested.

check_addon.py
```diff
@@ -103,6 +107,7 @@ def _check_for_invalid_xml_files(error_counter, file_index):
                 # Just try if we can successfully parse it
                 xml.etree.ElementTree.parse(xml_path)
             except xml.etree.ElementTree.ParseError:
+                xml_path = relative_path(xml_path)
```
Rechi: variable shouldn't be overwritten, either create a new variable, or do the `relative_path` function call directly in the log call

Rechi:
@mzfr please squash your changes into a single commit.

mzfr:
@Rechi I have made the changes and have squashed the commits

Rechi:
Tried it now, looks good
Only thing I would change is currently it outputs `/<ADDON-ID>/<FILE>`, but I don't like the slash at the beginning.

Razzeee:
@Rechi 
to be clear do you say you want no slash at all or `./` or `../`

Rechi:
no slash at the beginning (as it would look like an absolute path)
`./` is also okay for me, if you prefer that

Razzeee:
I would in deed prefer `./` as it communicates, that we maybe omited something

mzfr:
@Razzeee  @Rechi I have made the changes suggested.

Razzeee:
Thanks @mzfr

Issue #45, opened 2018-03-01
Think about creating a pip package from this


Razzeee:
I have never done a pip package so far, but it might work for us if we can use pipenv to run this.

slgobinath:
Hi,
Currently, I am working on this. Do you have a preferred universal name for addon-checker?
The name must be unique across Python packages available in [pypi](https://pypi.python.org/pypi) and can be used as the terminal command.

WDYT about `kodi-addon-checker`?

Razzeee:
Name sounds fine.
Code also looks good, but I've got no time right now to try it out. From my understanding that's what we need to push it to pypi?
Btw, we will have to change the author and email in the long run.

slgobinath:
Sure.
Please let me know the author name and email to use. I will update them and send a PR after testing on Windows.

FYI: Following this [tutorial](https://www.codementor.io/arpitbhayani/host-your-python-package-using-github-on-pypi-du107t7ku) I published my personal project to pypi.

Razzeee:
Please use `Team Kodi` as author, email is mandatory right?
We ultimatly want to push that package to pypi as Kodi with a Kodi account I guess.
See https://github.com/xbmc/addon-check/issues/49 please

slgobinath:
Author name and email are not mandatory. So I ignore the email but include `Team Kodi` as the author.
License field also skipped until we pick a license.

slgobinath:
I have tested setup.py installation on Linux Mint and Windows. Everything works as expected in my testing.
Please see the PR #50 

Thanks

mzfr:
@Razzeee since we now have a pip package, All thanks to @slgobinath. I think we can close this issue now.

PR #50, opened 2018-03-21
Standardize Python package

slgobinath:
This PR standardizes addon-check according to Python package format. With this PR, users can install `kodi-addon-checker` using `python setup.py install` or `pip install kodi-addon-checker` after publishing it to [PyPi](https://pypi.python.org/pypi)

Related issue: https://github.com/xbmc/addon-check/issues/45

Tested on Linux Mint 18.3 (Python 3.5.2)  and Windows 10 (Python 3.6.2)


Rechi:
The command is called `kodi-addon-checker`, but what it actually does is check a whole repo.
Can this be improved to have two commands (check repo & check addon)?

slgobinath:
In this PR I did not want to break the consistency. My plan was automating repo/addon folder detection.

For example, if the user is in a repo folder, it should check the complete repository. If the user is in addon folder, it should check for the addon. However, I postponed this to implement with command line interface (which will be my next PR).

Also, I will change the `kodi-addon-checker` to receive the folder to scan as an argument rather than using the current directory.
WDYT?

slgobinath:
I agree with you on executing from the current directory.

However, I find having two executable files is redundant (`kodi-addon-checker` and `kodi-repo-checker`)
I find two problems wit this:
1. In Windows, Python scripts are not executable. So we need two separate modules to provide `py -m ....` kind of execution. (I am not a Windows user. Please correct me if I am wrong)
2. Users need to choose which one to execute (little usability overhead)

WDYT about automatic folder detection?
If we find `addon.xml` in the current directory, treat it as an addon, if not treat it as a repo. If you prefer, we can print the mode at the beginning.
Eg: `Found xxxxxxxx plugin` or `Found xxxxxxxx repository`

As an alternative, we can ask users to configure it using command line argument but it will require the user to pick an option which is again a usability overhead.

mzfr:
I too feel that we shouldn't have two separate executables.

BTW, the current code behavior is:

1. If no path is passed, then run in "repo mode" - treating all top level folders in the current directory as separate addons.
2. If one or more paths are passed then:
    2.1 Read config file from current directory 
    2.2 Treat all paths as separate addons and check all of them

Maybe it's just me, but I've grown used to this model (of passing multiple paths etc.)

So, if we change this to detect at runtime whether a path is a repo/addon then we'll have to make sure we that we also allow for multiple paths.

romanvm:
> In Windows, Python scripts are not executable. So we need two separate modules to provide py -m .... kind of execution. (I am not a Windows user. Please correct me if I am wrong)

Not quite so. The current setuptools automatically bundles an .exe file for the script to be launched, that is,
```python
entry_points={'console_scripts': ['kodi-addon-checker = kodi_addon_checker.__main__:main']}
```
will produce `kodi-addon-checker.exe` in Scripts folder.

Razzeee:
If we can safely find a way to autodetect it, I'm all for it, as it lowers the bar for useage.

Still it would probably be best for our CI to pass multiple paths directly, for addon checks.

Razzeee:
API looks good, would be helpful if the help text points out the difference of repo vs addon check. Haven't looked at the code yet.

slgobinath:
New help description:
```
usage: kodi-addon-checker [-h] [--version] [add-on [add-on ...]]

Checks Kodi repo for best practices and creates problem and warning reports.
If optional add-on directories are provided, check only those add-ons.
Otherwise, scan current repository and check all add-ons in the current
directory.

positional arguments:
  add-on      optional add-on directories

optional arguments:
  -h, --help  show this help message and exit
  --version   show program's version number and exit
```
Please let me know if the wordings have to be improved.

Now add-on directory arguments are also checked for addon.xml file. If it does not exist, argparse will throw an error.

slgobinath:
Hi,
Please check my new commits.

> I just need to check what the repos will do when they automatically get the new version

Do you mean the `kodi-addon-checker` version in PyPi?

Razzeee:
No, as we don't have that yet.
What we do at the moment is update via git and call the py file. Which will not work anymore I think or maybe we can provide an "executable" py file.
For e.g. https://github.com/xbmc/repo-scripts/blob/helix/.travis.yml

slgobinath:
For this problem, I recommend publishing kodi-addon-checker to [PyPi](https://pypi.python.org/pypi), then install it using `pip install kodi-addon-checker` and access it as a command line tool.

```
...
script:
  - 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then kodi-addon-checker; fi'
  - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then kodi-addon-checker $(git diff --diff-filter=d --name-only HEAD~ | grep / | cut -d / -f1 | sort | uniq); fi'
...
```

However, there is a quick fix as well (Tested in my computer):
1. Include `.tests` directory to `PYTHONPATH`
    ```
    export PYTHONPATH="$PYTHONPATH:./.tests"
    ```
2. Run add-on checker as a Python module
    ```
    python -m kodi_addon_checker
    ```

Razzeee:
Pushed to pypi
https://pypi.org/project/kodi-addon-checker/
Seems like it does not like markdown? Upload guide said to try https://github.com/pypa/readme_renderer first, which I did not do.

slgobinath:
Thanks. I will fix the markdown issue.

slgobinath:
Sure thing. I will do that after #64 because it will help us to have separate reports per repository.

PR #51, opened 2018-03-21
Add check for dependencies within addon.xml

### LINKED ISSUE #29: Check for dependency issues
Maybe it's possible to check the included dependencies in addon.xml against the repo it's PRed to.
For this you could parse the repo addons.xml (or first download/unzip the addons.xml.gz)

mzfr:
Fixes #29 

I have tested it on [Twitch-on-kodi](https://github.com/MrSprigster/Twitch-on-Kodi/blob/master/addon.xml) addon.xml.
```
{'xbmc.python': '2.20.0', 'script.module.python.twitch': '2.0.0', 'script.module.requests': '2.9.1'}
Unrecognized requirement: xbmc.python 
Unrecognized requirement: script.module.python.twitch 
Version mismatch for addon script.module.requests. Required: 2.9.1, Available: 2.12.4 
```
Here `unrecognized requirements` means that these are not present in leia's addon.xml.
This is an issue because `xbmc.python` might be present in old branches and we are checking on leia only because PR was made on that.

Another issue is that in last line of the output we can see that for `script.module.requests` the available version is greater than the required version.

I have used leia's addon.xml for testing.

@Razzeee What should I do about these two issues ?


Rechi:
`xbmc.python` is in no `addons.xml` as it is shipped with Kodi and never needs to be updated (version describes xbmc interface and that can only change with a new Kodi version). The minimum versions for all Kodi versions can be found at https://github.com/xbmc/repo-scripts/blob/master/targets.cfg.

`version` specified in `import` tags describe min version, so use `LooseVersion` from `distutils` to compare them. The same is used in the repo generator to order versions.

mzfr:
@Rechi what about the packages like `script.module.python.twitch`. They might be present in old/other branches. 

Thanks for suggesting `distutils` I'll use it.

Rechi:
Kodi only uses one `addons.xml` (http://mirrors.kodi.tv/addons/<VERSION-CODE>/addons.xml). If addon isn't listed there it is not available for that version.

check_addon.py
```diff
@@ -349,3 +351,48 @@ def _logWarning(error_counter, warning_string):
...
+    content = requests.get(xml_url).content
+    tree = ET.fromstring(content).getroot()
+
+    return {
+        a.get("id"): a.get("version")
+        for a in tree.findall("addon")
+    }
+
+
+def get_users_dependencies(xml_file):
```
Razzeee: I think we can find a better name for this, as it has nothing to do with a `user`

check_addon.py
```diff
@@ -349,3 +351,48 @@ def _logWarning(error_counter, warning_string):
...
+    tree = ET.parse(xml_file).getroot()
+
+    return {
+        i.get("addon"): i.get("version")
+        for i in tree.findall("requires/import")
+    }
+
+def check_dependencies(error_counter, addon_xml):
+
+    branch_url = ROOT_URL.format(branch=os.environ.get("TRAVIS_BRANCH"))
```
Razzeee: This will cause this to only work on CI, it should be possible to run locally to. Probably after setting some variables.

mzfr: should I make it usable via `.tests-config.json` ?

Razzeee: Yes, do that. We might have to rethink that config file in the long run. But not now :)

mzfr: @Razzeee I have made it usable via .tests-config.json

kodi_addon_checker/check_addon.py
```diff
@@ -355,3 +361,49 @@ def _logWarning(error_counter, warning_string):
...
+    deps = get_users_dependencies(addon_xml)
+
+    for required_addon, required_version in deps.items():
+
+        if required_addon not in repo_addons:
+            error_counter = _logProblem(error_counter, "Unrecognized requirement: %s " % required_addon)
+
+        else:
+            available_version = repo_addons.get(required_addon)
+            if LooseVersion(available_version) != LooseVersion(required_version):
```
Rechi: `LooseVersion(available_version) < LooseVersion(required_version)`

mzfr:
@Razzeee Is there anything else I am supposed to do in it ?

Razzeee:
No, I just need to find some time/motivation to look at this. Hopfuly on friday at the latest. Sorry, for keeping you waiting.

kodi_addon_checker/check_addon.py
```diff
@@ -351,3 +357,49 @@ def _logWarning(error_counter, warning_string):
...
+
+    tree = ET.parse(addon_xml_path).getroot()
+
+    return {
+        i.get("addon"): i.get("version")
+        for i in tree.findall("requires/import")
+    }
+
+
+def check_dependencies(error_counter, addon_xml):
```
Rechi: `return error_counter` is missing for this function

PR #57, opened 2018-03-28
Add check for existing addon

### LINKED ISSUE #56: Check if addon is already present in a lower repository
This issue is a suggestion.

I have seen many cases in which a user submits an addon to a repository (lets say Krypton) and the addon is already present in "lower" repositories (e.g. helix). In those cases the check-tool should produce a warning to inform the user of this fact since most of the time the user is either not aware of the presence of the addon in a lower repo, doesn't know recent versions still have access to those lower repos, the changes are so minimal that do not justify a full addon review or the changes are not version specific. 
Finding if the addon already exists in another repo is both a memory exercise or a manual checking step (time consuming) and preferably should be avoided.

mzfr:
Fix #56 
@Razzeee This contain some code similar to PR [51](https://github.com/xbmc/addon-check/pull/51).

Should I make this also usable via `.tests-config.json` ?

### DIFF
--- a/kodi_addon_checker/__main__.py
+++ b/kodi_addon_checker/__main__.py
@@ -10,8 +10,6 @@ from kodi_addon_checker.record import Record, PROBLEM, WARNING, INFORMATION
 from kodi_addon_checker.report import Report
 from kodi_addon_checker.reporter import ReportManager
 
-ROOT_URL = "http://mirrors.kodi.tv/addons/{branch}/addons.xml"
-
 
 def dir_type(dir_path):
     """ArgParse callable to validate positional add-on arguments
@@ -38,16 +36,14 @@ def check_artifact(artifact_path, args, branch_name):
     :param args: argparse object
     :return: report
     """
-    branch_url = ROOT_URL.format(branch=branch_name)
-    repo_addons = check_addon._get_addons(branch_url)
 
     artifact_path = os.path.abspath(artifact_path)
     config = Config(artifact_path, args)
     ConfigManager.process_config(config)
     if os.path.isfile(os.path.join(artifact_path, "addon.xml")):
-        return check_addon.start(artifact_path, repo_addons, config)
+        return check_addon.start(artifact_path, branch_name, config)
     else:
-        return check_repo(artifact_path, config, repo_addons)
+        return check_repo(artifact_path, branch_name, config)
 
 
 def main():
--- a/kodi_addon_checker/__main__.py
+++ b/kodi_addon_checker/__main__.py
@@ -10,6 +10,19 @@ from kodi_addon_checker.record import Record, PROBLEM, WARNING, INFORMATION
 from kodi_addon_checker.report import Report
 from kodi_addon_checker.reporter import ReportManager
 
+ROOT_URL = "http://mirrors.kodi.tv/addons/{branch}/addons.xml"
+
+
+def _all_repo_addons():
+    branches = ['gotham', 'helix', 'isengard', 'jarvis', 'krypton', 'leia']
+    repo_addons = {}
+
+    for branch in branches:
+        branch_url = ROOT_URL.format(branch=branch)
+        repo_addons[branch] = check_addon._get_addons(branch_url)
+
+    return repo_addons
+
 
 def dir_type(dir_path):
     """ArgParse callable to validate positional add-on arguments
@@ -36,14 +49,14 @@ def check_artifact(artifact_path, args, branch_name):
     :param args: argparse object
     :return: report
     """
-
+    all_repo_addons = _all_repo_addons()
     artifact_path = os.path.abspath(artifact_path)
     config = Config(artifact_path, args)
     ConfigManager.process_config(config)
     if os.path.isfile(os.path.join(artifact_path, "addon.xml")):
-        return check_addon.start(artifact_path, branch_name, config)
+        return check_addon.start(artifact_path, branch_name, all_repo_addons, config)
     else:
-        return check_repo(artifact_path, branch_name, config)
+        return check_repo(artifact_path, branch_name, all_repo_addons, config)
 
 
 def main():
--- a/kodi_addon_checker/__main__.py
+++ b/kodi_addon_checker/__main__.py
@@ -49,7 +49,10 @@ def check_artifact(artifact_path, args, branch_name):
     :param args: argparse object
     :return: report
     """
+    logger = logging.getLogger(__package__)
+    logger.info("Downloading all repo addon list")
     all_repo_addons = _all_repo_addons()
+    logger.info("Download completed")
     artifact_path = os.path.abspath(artifact_path)
     config = Config(artifact_path, args)
     ConfigManager.process_config(config)
--- a/kodi_addon_checker/__main__.py
+++ b/kodi_addon_checker/__main__.py
@@ -10,7 +10,7 @@ from kodi_addon_checker.record import Record, PROBLEM, WARNING, INFORMATION
 from kodi_addon_checker.report import Report
 from kodi_addon_checker.reporter import ReportManager
 
-ROOT_URL = "http://mirrors.kodi.tv/addons/{branch}/addons.xml"
+ROOT_URL = "http://mirrors.kodi.tv/addons/{branch}/addons.xml.gz"
 
 
 def _all_repo_addons():
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -82,8 +82,8 @@ def start(addon_path, repo_addons, config=None):
 
             _check_for_invalid_xml_files(addon_report, file_index)
 
-            if config.is_enabled("check_for_existing_addon"):
-                check_for_existing_addon(addon_report, addon_path)
+            # if config.is_enabled("check_for_existing_addon"):
+            _check_for_existing_addon(addon_report, addon_path, repo_addons)
 
             _check_for_invalid_json_files(addon_report, file_index)
 
@@ -330,52 +330,6 @@ def _check_file_whitelist(report: Report, file_index, addon_path):
                                   relative_path(os.path.join(file["path"], file["name"]))))
 
 
-def get_repo_addons(branch):
-    branch_url = ROOT_URL.format(branch=branch)
-    content = requests.get(branch_url).content
-    tree = ET.fromstring(content)
-
-    return {
-        a.get("id"): (a.get("version"), bool(a.find("extension/broken")))
-        for a in tree.findall("addon")
-    }
-
-
-def get_addon_name(xml_path):
-    tree = ET.parse(xml_path).getroot()
-    return (tree.get("id"), tree.get("version"))
-
-
-def _check_for_existing_addon(error_counter, addon_xml):
-    addon_xml = os.path.join(addon_xml, "addon.xml")
-
-    branches = ['gotham', 'helix', 'isengard', 'jarvis', 'krypton', 'leia']
-
-    current_branch = os.environ.get("TRAVIS_BRANCH")
-
-    addon_name, addon_version = get_addon_name(addon_xml)
-
-    for branch in branches:
-
-        if branch >= current_branch:
-            break
-
-        else:
-            repo_addons = get_repo_addons(branch)
-
-            if addon_name not in repo_addons:
-                error_counter = _logWarning(error_counter, "This is a new addon")
-            else:
-                error_counter = _logWarning(error_counter, "%s addon already exist in %s branch with version %s."
-                                            % (addon_name, branch, repo_addons[addon_name][0]))
-                if repo_addons[addon_name][1] is False:
-                    error_counter = _logWarning(error_counter, "%s is broken" % addon_name)
-
-                break
-
-    return error_counter
-
-
 def relative_path(file_path):
     path_to_print = file_path[len(REL_PATH):]
     return ".{}".format(path_to_print)
@@ -404,6 +358,7 @@ def _check_complex_addon_entrypoint(report: Report, addon_path, max_entrypoint_l
 
 
 def number_of_lines(report: Report, filepath: str, library: str, max_entrypoint_line_count: int):
+
     with open(filepath, 'r') as file:
         data = file.read()
 
@@ -426,9 +381,7 @@ def number_of_lines(report: Report, filepath: str, library: str, max_entrypoint_
 
 
 def _get_addons(xml_url):
-    """
-        addon.xml for the target Kodi version
-    """
+    """addon.xml for the target Kodi version"""
     content = requests.get(xml_url).content
     tree = ET.fromstring(content)
 
@@ -453,6 +406,7 @@ def _get_users_dependencies(addon_path):
 
 
 def _check_dependencies(report: Report, addon_path, repo_addons):
+    """Check for any new dependencies in addon.xml file"""
     deps = _get_users_dependencies(addon_path)
     ignore = ['xbmc.json', 'xbmc.gui', 'xbmc.json',
               'xbmc.metadata', 'xbmc.python']
@@ -474,3 +428,23 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
             elif LooseVersion(available_version) < LooseVersion(required_version) and (required_addon not in ignore):
                 report.add(Record(PROBLEM, "Version mismatch for addon %s. Required: %s, Available: %s "
                                   % (required_addon, required_version, available_version)))
+
+
+def get_addon_name(xml_path):
+    tree = ET.parse(xml_path).getroot()
+    return (tree.get("id"), tree.get("version"))
+
+
+def _check_for_existing_addon(report: Report, addon_xml, repo_addons):
+    """Check if addon submitted already exists or not"""
+    addon_xml = os.path.join(addon_xml, "addon.xml")
+    addon_name, addon_version = get_addon_name(addon_xml)
+
+    if addon_name not in repo_addons:
+        report.add(Record(WARNING, "This is a new addon"))
+    else:
+        report.add(Record(WARNING, "%s addon already exist with version %s."
+                          % (addon_name, repo_addons[addon_name][0])))
+
+        if repo_addons[addon_name][1] is False:
+            report.add(Record(WARNING, "%s is broken" % addon_name))
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -82,6 +82,9 @@ def start(addon_path, repo_addons, config=None):
 
             _check_for_invalid_xml_files(addon_report, file_index)
 
+            if config.is_enabled("check_for_existing_addon"):
+                check_for_existing_addon(addon_report, addon_path)
+
             _check_for_invalid_json_files(addon_report, file_index)
 
             _check_artwork(addon_report, addon_path, addon_xml, file_index)
@@ -337,8 +340,7 @@ def check_for_existing_addon(error_counter, addon_xml):
 
     current_branch = os.environ.get("TRAVIS_BRANCH")
 
-    branches = ['dharma', 'eden', 'frodo', 'gotham', 'helix',
-                'isengard', 'jarvis', 'krypton', 'leia']
+    branches = ['gotham', 'helix', 'isengard', 'jarvis', 'krypton', 'leia']
 
     addon_name, addon_version = get_addon_name(addon_xml)
 
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -15,7 +15,6 @@ from kodi_addon_checker.record import PROBLEM, Record, WARNING, INFORMATION
 from kodi_addon_checker.report import Report
 
 REL_PATH = ""
-ROOT_URL = "http://mirrors.kodi.tv/addons/{branch}/addons.xml"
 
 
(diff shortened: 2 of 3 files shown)


Razzeee:
Yes, our code should never directly access travis variables like `os.environ.get("TRAVIS_BRANCH")` does

check_addon.py
```diff
@@ -334,6 +338,48 @@ def _check_file_whitelist(error_counter, file_index, addon_path):
...
+    tree = ET.parse(xml_path).getroot()
+    return (tree.get("id"), tree.get("version"))
+
+
+def check_for_existing_addon(error_counter, addon_xml):
+    addon_xml = os.path.join(addon_xml, "addon.xml")
+
+    current_branch = os.environ.get("TRAVIS_BRANCH")
+
+    branches = ['dharma', 'eden', 'frodo', 'gotham', 'helix',
```
Rechi: don't check dharma, eden & frodo

Rechi:
Try to cache `addons.xml` files as much as possible.
With your current code a repo check will do `9 x number of addons` requests.

mzfr:
@Razzeee I have made it usable via .tests-config.json

Razzeee:
I would like another approach more, what your doing at the moment:
1. make sure nobody can run the test offline/without travis - so configure that test not to run

How about: 
2. Set a variable for `TRAVIS_BRANCH` in our config, pull it from there. That would solve the repo usecase. 

It might still be problematic for a dev to run, as he might not have the complete repo and the config file. So he would need to pass the target platform as param. Another approach might be to just extract that from the addon.xml. So we would check which kodi version is targeted there and just expect that to be the branch name. With that you could then do `os.environ.get("TRAVIS_BRANCH")`  see if it's not empty and check if it's equal to the one we got from the addon.xml

mzfr:
@Razzeee Currently we are using `os.environ.get("TRAVIS_BRANCH")` to get the branch name and version of the addon(submitted by developer). This ensures that we don't search for addons submitted to `helix` in any branch above it i.e `isengard` or `jarvis` etc.

If we want to do everything with config file then I think we should decide a good way(maybe syntax/format of it.) to use it as you said [here](https://github.com/xbmc/addon-check/pull/51#discussion_r176574710). 
yes if we are not going to have target in config file then you can run it by `TRAVIS_BRANCH=leia python3 check_repo <addon path>` to pass branch name via command line.

What I am able to understand from that last part is that we should get kodi version from addon.xml and then check whether we have got it right using or not(using `TRAVIS_BRANCH`) but isn't our current approach better then this I mean we are directly getting the branch name without having to do any checks for it. Am I missing something here ?

Also maybe we can depend on `TRAVIS_BRANCH` variable for what @Rechi said about caching the addon.xml. We can have a check that if current branch is different then the previous only then make a request for the addon.xml else use the old one. 
@Rechi what do you think about it ?

Razzeee:
Well what I'm trying to say is that I want this to also work locally on only one addon. And I don't think it's good design to have users set environment variables first or at all.

Razzeee:
1. I think the functionallity is fine, I just want it to be able to work without travis (locally for e.g.)
2. On the other hand, it might still be relevant information to know if an addon is available in a newer branch? @enen92

check_addon.py
```diff
@@ -334,6 +339,52 @@ def _check_file_whitelist(error_counter, file_index, addon_path):
...
+def _check_for_existing_addon(error_counter, addon_xml):
+    addon_xml = os.path.join(addon_xml, "addon.xml")
+
+    branches = ['gotham', 'helix', 'isengard', 'jarvis', 'krypton', 'leia']
+
+    current_branch = os.environ.get("TRAVIS_BRANCH")
+
+    addon_name, addon_version = get_addon_name(addon_xml)
+
+    for branch in branches:
```
Rechi: Actually you only have to check the previous version.
If it is present in any lower version it will be also present there.

check_addon.py
```diff
@@ -334,6 +339,52 @@ def _check_file_whitelist(error_counter, file_index, addon_path):
...
+
+        if branch >= current_branch:
+            break
+
+        else:
+            repo_addons = get_repo_addons(branch)
+
+            if addon_name not in repo_addons:
+                error_counter = _logWarning(error_counter, "This is a new addon")
```
Rechi: With the current for loop over all branches, this might be wrong and/or logged multiple times for the same addon.

mzfr: Yeah, right. Should I just remove this check then? If it doesn't already exist in any branch then the addon is new.

mzfr:
@Razzeee I have made the required changes. please review this.

kodi_addon_checker/check_addon.py
```diff
@@ -435,3 +389,23 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+
+def _check_for_existing_addon(report: Report, addon_xml, repo_addons):
+    """Check if addon submitted already exists or not"""
+    addon_xml = os.path.join(addon_xml, "addon.xml")
+    addon_name, addon_version = get_addon_name(addon_xml)
+
+    if addon_name not in repo_addons:
+        report.add(Record(WARNING, "This is a new addon"))
+    else:
+        report.add(Record(WARNING, "%s addon already exist with version %s."
```
mzfr: Notice here we are not printing the branch name that is because we get the repo addon directly from the __main__ file, that was done to cache the addon.xml file.

kodi_addon_checker/check_addon.py
```diff
@@ -14,6 +14,7 @@
 from kodi_addon_checker.report import Report
 
 REL_PATH = ""
+ROOT_URL = "http://mirrors.kodi.tv/addons/{branch}/addons.xml"
```
Rechi: unused

kodi_addon_checker/check_addon.py
Rechi: This shouldn't be a warning (only info), as it only checks if the addon is already in the repo which it is PRed to.

Rechi:
As always, at least `Resolved conflicts and removed duplicate code` commit has to be squashed.

kodi_addon_checker/check_addon.py
```diff
@@ -384,3 +392,23 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+
+
+def _check_for_existing_addon(report: Report, addon_xml, repo_addons, branch_name):
+    """Check if addon submitted already exists or not"""
+    addon_xml = os.path.join(addon_xml, "addon.xml")
+    addon_name, addon_version = get_addon_name(addon_xml)
+
+    if addon_name not in repo_addons:
+        report.add(Record(INFORMATION, "This is a new addon"))
+    else:
```
Razzeee: We need another case here if the addon is in repo_addon and has the same version there as it has in the local addon.xml. In that case you need to Record a PROBLEM as we can't accept addons that are not version bumped.

kodi_addon_checker/check_addon.py
```diff
@@ -384,3 +392,23 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+        report.add(Record(INFORMATION, "%s addon already exist in branch %s with version %s."
+                          % (addon_name, branch_name, repo_addons[addon_name][0])))
+
+        if repo_addons[addon_name][1] is False:
```
Razzeee: Can we write this in a better way? I would guess it's trying to get the broken flag, due to the code below, but it's not very obvious.

kodi_addon_checker/check_addon.py
```diff
@@ -384,3 +392,23 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+            report.add(Record(WARNING, "%s is broken" % addon_name))
```
Razzeee: Make it `"%s has previously been marked as broken."`

kodi_addon_checker/check_addon.py
```diff
@@ -79,6 +83,9 @@ def start(addon_path, repo_addons, config=None):
 
             _check_for_invalid_xml_files(addon_report, file_index)
 
+            if config.is_enabled("check_for_existing_addon"):
```
Razzeee: I think we should be able to drop this check. As we would try to get that config file down to zero. Having both a config file and command line parameters might be very confusing in the long run.

kodi_addon_checker/check_addon.py
```diff
@@ -399,16 +398,29 @@ def get_addon_name(xml_path):
...
-    if addon_name not in repo_addons:
-        report.add(Record(INFORMATION, "This is a new addon"))
-    else:
-        report.add(Record(INFORMATION, "%s addon already exist in branch %s with version %s."
-                          % (addon_name, branch_name, repo_addons[addon_name][0])))
+    branches = ['gotham', 'helix', 'isengard', 'jarvis', 'krypton', 'leia']
+
+    for branch in branches:
+
+        if branch >= addon_branch:
```
Razzeee: I don't think we need this. We always want to see the else branch. Even if I commit a addon to a highter branch it's useful to know that we once had an ddon with the same name. Or we are clashing with aversion present in an old branch.

mzfr: I have fixed this.

kodi_addon_checker/check_addon.py
```diff
@@ -384,3 +391,33 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+        branch_url = ROOT_URL.format(branch=branch)
+        repo_addons = _get_addons(branch_url)
+
+        if addon_name in repo_addons:
+            if LooseVersion(addon_version) == LooseVersion(repo_addons[addon_name]):
+                report.add(Record(PROBLEM, "%s addon already exist with current version in lower branch"
+                                  % addon_name))
+
+            else:
+                report.add(Record(INFORMATION, "%s addon also exists in %s branch but with lower version"
```
Rechi: as you compare for equality in the if statement the addon repo version could be higher too

kodi_addon_checker/check_addon.py
```diff
@@ -384,3 +391,33 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+    """Check if addon submitted already exists or not"""
+    addon_xml = os.path.join(addon_path, "addon.xml")
+    addon_name, addon_version = get_addon_name(addon_xml)
+
+    branches = ['gotham', 'helix', 'isengard', 'jarvis', 'krypton', 'leia']
+
+    for branch in branches:
+
+        branch_url = ROOT_URL.format(branch=branch)
+        repo_addons = _get_addons(branch_url)
```
Rechi: Now you have reintroduced the code which does 6 request for each addon in a repo. I don't care about the 6 request, but about downloading those files multiple times.

kodi_addon_checker/check_addon.py
```diff
@@ -384,3 +391,33 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+            if LooseVersion(addon_version) == LooseVersion(repo_addons[addon_name]):
+                report.add(Record(PROBLEM, "%s addon already exist with current version in lower branch"
+                                  % addon_name))
+
+            else:
+                report.add(Record(INFORMATION, "%s addon also exists in %s branch but with lower version"
+                                  % (addon_name, branch)))
+                break
+        else:
+            report.add(Record(INFORMATION, "This is a new addon"))
```
Rechi: This will report multiple times and may be even report a wrong state.
E.g addon not in gotham and helix, but in isengard -> 2 times `This is a new addon` and then addon exists already

kodi_addon_checker/check_addon.py
mzfr: oh instead of having `==` I should have `<=` so if the repo_version will be higher or equal then it will give `addon already exist with current version in lower branch` right ?

kodi_addon_checker/check_addon.py
mzfr: you mean when we will run this on whole repo right ?

mzfr: you mean for whole repo check ?

kodi_addon_checker/check_addon.py
mzfr: yes thanks for pointing that out.

kodi_addon_checker/check_addon.py
Rechi: Yes for a whole repo check you this will download the same 6 files as many times as the repo has addons.
Notice: In most cases not exactly 6 x addons as the for loop has a break

kodi_addon_checker/check_addon.py
mzfr: yes thanks for pointing that out.

solution I can think off right now is to have a counter, initializing it with 6 i.e `branch_count=6` and instead of putting `       report.add(Record(INFORMATION, "This is a new addon"))` in else we can do `branch_count-1` and then check whether `branch_count==0` or not, outside the loop.

Rechi: simpler solution: move the report after the for loop, replace break with return

kodi_addon_checker/check_addon.py
mzfr: ok so maybe we can add a new function that make total of 6 reqeuests and will return a dictionary. That will have branch name as `key` and for `value` we will have another dictionary that will have respective repo_addons 
so it would look like:

```
{gotham:{<all the addon list in gotham>},'helix':{<all the addon list in helix>}}
```
Then we can use that dictionary again and again without having to make more than 6 requests

mzfr:
@Rechi I have put a new function `_all_repo_addons` that will solve the issue. Now we make exact 6 requests to get all the repo_addon list.

please review this.

kodi_addon_checker/check_addon.py
```diff
@@ -384,3 +389,31 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+
+
+def _check_for_existing_addon(report: Report, addon_path, all_repo_addons):
+    """Check if addon submitted already exists or not"""
+    addon_xml = os.path.join(addon_path, "addon.xml")
+    addon_name, addon_version = get_addon_name(addon_xml)
+
+    branches = ['gotham', 'helix', 'isengard', 'jarvis', 'krypton', 'leia']
+
+    for branch in branches:
```
Rechi: `for branch in branches`

kodi_addon_checker/check_addon.py
```diff
@@ -384,3 +389,31 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+        repo_addons = all_repo_addons[branch]
+
+        if addon_name in repo_addons:
+            if LooseVersion(addon_version) <= LooseVersion(repo_addons[addon_name]):
+                report.add(Record(PROBLEM, "%s addon already exist with current version in lower branch"
```
Rechi: I would change 'exist with **current version** in **lower branch**' to something mentioning the version and the branch

kodi_addon_checker/check_addon.py
```diff
@@ -384,3 +389,31 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+                                  % addon_name))
+                return
+            else:
+                report.add(Record(INFORMATION, "%s addon also exists in %s branch but with lower version"
```
Rechi: mention the version

kodi_addon_checker/check_addon.py
mzfr: sorry I don't understand this ?

Rechi: sorry wrong variable
`for branch in all_repo_addons:` and get rid of the now unused variable branches
defining branches 2 times is error prone

mzfr: ok but then for that I'll have to reverse the order of branches in the `branches` list so that gotham is the first branch and leia is the last one because then if krypton will be the last one added to the dictionary then we'll have different output

if `for branch in branches:` then
```python
INFO: Checking add-on plugin.video.twitch
INFO: Created by A Talented Community
INFO: Addon id matches folder name
**INFO: plugin.video.twitch addon also exists in  gotham branch but with version: 1.4.1**
INFO: Image icon exists
INFO: icon dimensions are fine 512x512
INFO: Image fanart exists
INFO: fanart dimensions are fine 3840x2160
WARN: Complex entry point. Check: resources/lib/routes.py | Counted lines: 1286 | Lines allowed: 80
WARN: Complex entry point. Check: resources/lib/service.py | Counted lines: 155 | Lines allowed: 80
WARN: We found no problems and 2 warnings, please check the logfile.
```
--------------------------------------------------------------------------
if `for branch in all_repo_addons:` then
```python
INFO: Checking add-on plugin.video.twitch
INFO: Created by A Talented Community
INFO: Addon id matches folder name
**ERROR: plugin.video.twitch addon already exist with version: 2.2.2 in krypton branch**
INFO: Image icon exists
INFO: icon dimensions are fine 512x512
INFO: Image fanart exists
INFO: fanart dimensions are fine 3840x2160
WARN: Complex entry point. Check: resources/lib/routes.py | Counted lines: 1286 | Lines allowed: 80
WARN: Complex entry point. Check: resources/lib/service.py | Counted lines: 155 | Lines allowed: 80
ERROR: We found 1 problems and 2 warnings, please check the logfile.
```
Notice the difference in the `Error` and `Info` message

mzfr:
@Razzeee I have made the changes suggested by @Rechi

kodi_addon_checker/check_addon.py
```diff
@@ -384,3 +389,29 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
             if LooseVersion(available_version) < LooseVersion(required_version) and (required_addon not in ignore):
                 report.add(Record(PROBLEM, "Version mismatch for addon %s. Required: %s, Available: %s "
                                   % (required_addon, required_version, available_version)))
+
+
+def get_addon_name(xml_path):
```
Razzeee: This is private, so please prefix it accordingly with `_`

kodi_addon_checker/__main__.py
```diff
@@ -11,6 +11,17 @@
 ROOT_URL = "http://mirrors.kodi.tv/addons/{branch}/addons.xml"
 
 
+def _all_repo_addons():
+    branches = ['gotham', 'helix', 'isengard', 'jarvis', 'krypton', 'leia']
+    repo_addons = {}
+
+    for branch in branches:
+        branch_url = ROOT_URL.format(branch=branch)
+        repo_addons[branch] = check_addon._get_addons(branch_url)
```
Razzeee: Please add some logging here and portentially in _get_addons. Locally it seems to get stuck on downloading or well take forever. Might be my connection. But logging something might not hurt. 
Do normal logs, not our report stuff :)

Razzeee: Or well don't do that as it will likely show up in the travis log too and mess up our reports.
Kinda complicated to have both.

Rechi: maybe load the addons.xml.gz version, it is much smaller

mzfr: Then we'll have to unzip it right ?

mzfr: yes we can do that. we'll have to add something like this:
```python
  import gzip
  with gzip.open('file.tar.gz', 'rb') as f:
  file_content = f.read()
```
Then add the `file_content` to the dictionary.
Should I do it ?

Razzeee: Absolutely should be an awesome improvement. I'm far more concerned for a file transfer to break then a local zip operation not working :)

mzfr:
@Razzeee I have made the changes. 
I have changed `get_addon_name` to `_get_addon_name` in first commit and second commit have all the other changes

Razzeee:
Did you runtime test and time it? Can you post the timings?

mzfr:
yes I did.
here are the times:
```
24.647593220001
21.481202986000426
18.47452666100071
21.899181321998185
```

Razzeee:
Will try to test this tonight. Kinda disappointed by those timings, hoped for a bigger improvement.

mzfr:
sure !! and let me know if any changes are required :grin:

Razzeee:
Can we timeout and retry on the requests please? If that doesn't help please abort and report an error.

mzfr:
@Razzeee I have made the changes

kodi_addon_checker/check_addon.py
```diff
@@ -8,13 +8,17 @@
...
+from io import BytesIO
+
 from PIL import Image
 
 from kodi_addon_checker.common import has_transparency
 from kodi_addon_checker.record import PROBLEM, Record, WARNING, INFORMATION
 from kodi_addon_checker.report import Report
 
 REL_PATH = ""
+logger = logging.getLogger(__name__)
```
Razzeee: Uppercase this to make clear it's a global

kodi_addon_checker/check_addon.py
```diff
@@ -424,3 +440,29 @@ def _check_dependencies(report: Report, addon_path, repo_addons):
...
+    for branch in sorted(all_repo_addons):
+        repo_addons = all_repo_addons[branch]
+
+        if addon_name in repo_addons:
+            if LooseVersion(addon_version) <= LooseVersion(repo_addons[addon_name]):
+                report.add(Record(PROBLEM, "%s addon already exist with version: %s in %s branch"
+                                  % (addon_name, repo_addons[addon_name], branch)))
+                return
+            else:
+                report.add(Record(INFORMATION, "%s addon also exists in %s branch but with version: %s"
```
Razzeee: Not sure if this is good or just confusing, we might want to remove this later

PR #58, opened 2018-03-30
Update travis examples and short summary

Razzeee:
Check addon example is not tested at the moment.

### DIFF
--- a/.travis-addon.yml.example
+++ b/.travis-addon.yml.example
@@ -3,12 +3,8 @@ python:
   - "3.6"
 
 install:
-  - pip install Pillow requests radon
-
-before_install:
-  - git submodule update --init --recursive --remote
+  - pip install kodi-addon-checker
 
 # command to run our tests
-script:
-  - cd .tests
-  - python check_addon.py
+script: 
+  - kodi-addon-checker .
--- a/.travis-repo.yml.example
+++ b/.travis-repo.yml.example
@@ -3,16 +3,12 @@ python:
   - "3.6"
 
 install:
-  - pip install Pillow requests radon
-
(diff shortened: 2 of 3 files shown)


.travis-addon.yml.example
```diff
@@ -3,12 +3,8 @@ python:
...
-before_install:
-  - git submodule update --init --recursive --remote
+  - pip install Pillow requests kodi-addon-checker
 
 # command to run our tests
 script: 
-  - cd .tests 
-  - python check_addon.py
\ No newline at end of file
+  - kodi-addon-checker .
```
slgobinath: To check an entire repository the `.` is not necessary.
Directories passed as arguments are treated as plugins (The same behavior as the previous implementation)

This line should be:
```
- kodi-addon-checker
```

.travis-addon.yml.example
```diff
@@ -3,12 +3,8 @@ python:
   - "3.6"
 
 install:
-  - pip install Pillow requests
-
-before_install:
-  - git submodule update --init --recursive --remote
+  - pip install Pillow requests kodi-addon-checker
```
slgobinath: Though it is perfectly fine to install Pillow and requests, it is not necessary to do so.

You can simply call `pip install kodi-addon-checker` which will install `kodi-addon-checker` along with `Pillow` and `requests` because they are already marked as dependencies in [setup.py#L5](https://github.com/xbmc/addon-check/blob/master/setup.py#L5)

Razzeee: makes sense, forgot about that when adding the new dependency

Razzeee:
Updated

PR #63, opened 2018-04-04
[WIP]Add check for dependencies within addon.xml

mzfr:
Sorry, I messed up #51 by merging the branches (via GitHub) into it instead of re-basing (I will keep that in mind) and when I later squashed the commits, **something** went wrong.

I am working on the changes @Rechi requested i.e
* request addons.xml only once (for a repo check)
* xbmc.(addon|gui|json|metadata|python) dependencies either have to be skipped, or somehow added as available


mzfr:
@Razzeee @Rechi I have added a check for ignore-ing the `xbmc.(addon|gui|json|metadata|python)`.
Also I have made the necessary changes for making single request for the whole repo.

Please review this and let me know if there's any changes required.

kodi_addon_checker/check_addon.py
```diff
@@ -351,3 +356,47 @@ def _logWarning(error_counter, warning_string):
...
+    }
+
+
+def _check_dependencies(error_counter, addon_path, repo_addons):
+    deps = get_users_dependencies(addon_path)
+    ignore = ['xbmc.json', 'xbmc.gui', 'xbmc.json', 'xbmc.metadata', 'xbmc.python']
+
+    for required_addon, required_version in deps.items():
+        if (required_addon not in repo_addons) and (required_addon not in ignore):
+            error_counter = _logProblem(error_counter, "Unrecognized requirement: %s " % required_addon)
```
Razzeee: I think we can word this better, how about `Required addon "%s" not available in current repository.`

kodi_addon_checker/check_addon.py
```diff
@@ -351,3 +356,47 @@ def _logWarning(error_counter, warning_string):
...
+
+        else:
+            available_version = repo_addons.get(required_addon)
+            if LooseVersion(available_version) != LooseVersion(required_version) and (required_addon not in ignore):
+                error_counter = _logWarning(error_counter, "Version mismatch for addon %s. Required: %s, Available: %s "
```
Razzeee: Might be worth to point to the addons `addon.xml` here, to make this clearer.

kodi_addon_checker/check_repo.py
```diff
@@ -22,6 +24,9 @@ def check_repo(repo_path, parameters):
     else:
         toplevel_folders = sorted(parameters)
 
+    branch_url = ROOT_URL.format(branch=os.environ.get("TRAVIS_BRANCH"))
```
Razzeee: Please don't use `os.environ.get("TRAVIS_BRANCH")` anywhere. Read it from the params or the config file. Probably both would be good.

kodi_addon_checker/check_repo.py
```diff
@@ -30,7 +35,7 @@ def check_repo(repo_path, parameters):
         if addon_folder[0] != '.':
             addon_path = os.path.join(repo_path, addon_folder)
             error_counter = check_addon.start(
-                error_counter, addon_path, config)
+                error_counter, addon_path, repo_addons, config)
```
Razzeee: So this will work for a whole repo review, but how does it work for a single addon?

kodi_addon_checker/check_repo.py
mzfr: ok so we can have a parameter `branch` that will have the same functionality as `TRAVIS_BRANCH`. And for default we can set it's value to None.

Razzeee: And what do you do on `None`?

mzfr: yeah that's what I was thinking. Maybe  just not run those check that are dependent on `branch` parameters or  maybe instead of having `None` as default we can have `leia` or  `krypton` as default (assuming most of the add-on will be submitted to those because they are new) ?

Razzeee: We could also read the addon.xml and parse the python version to figure out the branch? But a warning, this will only work for single addons. 

See this list https://kodi.wiki/view/Addon.xml#Dependency_versions, but as it turns out it's not a 100% as krypton and leia has the same target here.

Feels like we should make kodi version a needed param. That would also enable us to move most of .tests-config.json into addon-check and just map it to the versions.

mzfr: @Razzeee yes this can be done but I think instead of taking the kodi version as parameter and then changing it into a branch name (for getting addons.xml), we can just have a branch name as parameter.  This will save us some code for converting version into branch name.

And about  `.tests-config.json`, it currently doesn't have any code that actually depends on version.
 For example: this is a config file from branch `leia`
```json
{
    "check_legacy_language_path": true,
    "check_kodi_leia_deprecations": true
}
```
Almost all the branch have similar config file and here none of the function directly depends upon the version. So might not have a huge impact on config but yes it will be helpful in future.

Razzeee: sorry, that's what I ment by kodi version. As the should match branch names. SO we should be fine.

`.tests-config.json` is completly version dependent, what you state is wrong. It's pretty much the whole reason to even have it. The `check_kodi_leia_deprecations` option is pretty clearly refering to every branch leia+. Please look again and let me know if you have problems understanding :)

mzfr: ok so I'll just add an argument for it.

I had a look to what `check_kodi_leia_deprecation` and what I am able to understand is that it check for strings that were present in old add-ons but are not supposed to be in any add-on submitted to Leia right ?

Razzeee: To leia or newer, yes.
The other config items basically work along the same lines. Usually some version dependent changes to kodi engine.

mzfr:
@Razzeee I have implemented the changes please review this.
For some reason I didn't got the "report" changes when i was rebase-ing so I did them manually(sorry).
Also pycodestyle tests are failing due to the "escape sequences" but why are we using those escape sequences ?

mzfr:
@Razzeee I have rebased it. Now you can review this.

kodi_addon_checker/__main__.py
```diff
@@ -66,7 +67,7 @@ def main():
         for directory in args.dir:
             report.add(check_artifact(directory, args))
```
Razzeee: This should also honor the branch argument, shouldn't it?

mzfr: yes and actually I have also missed to add another argument in `check_artifact` function.

 But what I am now finding contradictory is that if we pass single addon  [this](https://github.com/xbmc/addon-check/blob/master/kodi_addon_checker/__main__.py#L41) will directly call the start function. But we are getting [`repo_addons`](https://github.com/xbmc/addon-check/pull/63/files#diff-87cbf8a463aa7f67c8382bdac08ec809R15) from `check_repo` function. So I'll have to do this within the start function but if I do that then we'll have the same issue of having multiple requests for same repo i.e caching addon.xml.

@Razzeee Am I on the right track or have I missed something here ?

Razzeee: That evaluation sounds right.

mzfr: I am currently trying to figure it out but I am open to suggestions :wink:

Razzeee: do a preparation step in __main__.py where you get everything that we want to keep in that overarching scope?
Alternative might be to even write that to the disk, not sure I'm a fan of that.

Razzeee: We'll you can always add another class between those. If theres a good reason to do this.

mzfr:
@Razzeee I have implemented the changes. Please review them.

Razzeee:
This is missing the installation of `requests` in `setup.py`

Razzeee:
You need to guard the parameter against not being set somehow. Otherwise we will try to get a branch with the name `None`

kodi_addon_checker/__main__.py
```diff
@@ -57,16 +61,18 @@ def main():
     parser.add_argument("--version", action="version",
                         version="%(prog)s 0.0.1")
     parser.add_argument("dir", type=dir_type, nargs="*", help="optional add-on or repo directories")
+    parser.add_argument("--branch", help="Branch name on which checker is to run")
```
Razzeee: This is required https://docs.python.org/2/library/argparse.html#required

Razzeee: I would also suggest to hardcode the choices to the available repos https://docs.python.org/2/library/argparse.html#choices

Razzeee:
```
    parser.add_argument("--branch", choices=['gotham', 'helix', 'isengard', 'jarvis', 'krypton', 'leia'],
                        required=True, help="Target branch name where the checker will resolve dependencies")
```

mzfr: I have made the suggested changes.

kodi_addon_checker/check_addon.py
```diff
@@ -72,6 +74,9 @@ def start(addon_path, config=None):
         if len(addon_xml.findall("*//broken")) == 0:
             file_index = _create_file_index(addon_path)
 
+            if config.is_enabled("check_dependencies"):
```
Razzeee: Please remove this setting. Should be not needed, now that the branch parameter is mandatory.

kodi_addon_checker/check_addon.py
```diff
@@ -302,3 +308,46 @@ def _check_file_whitelist(report: Report, file_index, addon_path):
...
+
+    for required_addon, required_version in deps.items():
+        if required_addon not in repo_addons:
+            if required_addon not in ignore:
+                report.add(Record(WARNING, "Required addon %s not available in current repository." % required_addon))
+        else:
+            available_version = repo_addons[required_addon]
+
+            if LooseVersion(available_version) != LooseVersion(required_version) and (required_addon not in ignore):
+                report.add(Record(WARNING, "Version mismatch for addon %s. Required: %s, Available: %s "
```
Razzeee: This should be a `PROBLEM` not a `WARNING`

kodi_addon_checker/check_addon.py
```diff
@@ -137,12 +142,13 @@ def _check_addon_xml(report: Report, addon_path):
...
-        addon_xml = xml.etree.ElementTree.parse(addon_xml_path)
+        addon_xml = ET.parse(addon_xml_path)
         addon = addon_xml.getroot()
         report.add(Record(INFORMATION, "Created by %s" % addon.attrib.get("provider-name")))
         _addon_xml_matches_folder(report, addon_path, addon_xml)
-    except xml.etree.ElementTree.ParseError:
-        report.add(Record(PROBLEM, "Addon xml not valid, check xml. %s" % relative_path(addon_xml_path)))
+
+    except ET.ParseError:
+        report.add(Record(INFORMATION, "Created by %s" % addon.attrib.get("provider-name")))
```
Razzeee: Is this a copy and paste error?

kodi_addon_checker/check_addon.py
mzfr: should [Line 347](https://github.com/xbmc/addon-check/pull/63/files#diff-0908daa4913eca95dda1221fb9159098R347) be a `PROBLEM` or should I leave it as a `WARNING
`

Razzeee: Yes, absolutely. Sorry for missing that.

kodi_addon_checker/check_addon.py
```diff
@@ -137,11 +141,12 @@ def _check_addon_xml(report: Report, addon_path):
     try:
         _addon_file_exists(report, addon_path, r"addon\.xml")
 
-        addon_xml = xml.etree.ElementTree.parse(addon_xml_path)
+        addon_xml = ET.parse(addon_xml_path)
         addon = addon_xml.getroot()
-        report.add(Record(INFORMATION, "Created by %s" % addon.attrib.get("provider-name")))
+        report.add(Record(PROBLEM, "Addon xml not valid, check xml. %s" % relative_path(addon_xml_path)))
```
Razzeee: The other one is okay now, but you changed this one now. I don't think you ment to do this.

mzfr: I am really sorry for this. I'll fix it right away.

Razzeee:
Needs a rebase now

mzfr:
@Razzeee I have rebased it now.

kodi_addon_checker/check_addon.py
```diff
@@ -337,3 +341,58 @@ def number_of_lines(filepath):
         data = file.read()
 
     return (analyze(data).lloc)
+
+
+def _logProblem(error_counter, problem_string):
```
Razzeee: I suspect a rebase error here. THis and the next method are not supposed to be used anymore.

PR #64, opened 2018-04-05
Improve addon report infrastructure

slgobinath:
Hi @Razzeee,
Could you please have a look at this design improvement. This PR improves how addon-checker handles reports. Currently, all reports are printed to the console as you check add-ons. It cannot support multiple reporters (say GitHub reporter, markdown reporter, checkstyle reporter, etc).

This PR isolates reports and reporters into separate modules so that add-on checker can be easily extended to support new reporters in future. For the moment, I have implemented [ConsoleReporter](https://github.com/slgobinath/addon-check/blob/feature-new-architecture/kodi_addon_checker/plugins/console_reporter.py) which is printing the report to the console as we have right now.

The low level [Record](https://github.com/slgobinath/addon-check/blob/feature-new-architecture/kodi_addon_checker/report.py#L29) object is designed to include line numbers and character level positions for future requirements (to report errors in Python scripts). This design also allows you to report problems and/or warnings at repo level, add-on level or file level.

I have also defined three log levels: 'INFO', 'WARN' and 'ERROR'(which is equivalent to 'PROBLEM'). These terms were chosen to make the output more structured.

Sample report generated according to this implementation:
```
Checking /home/gobinath/Workspace/slgobinath/kodi/repo-plugins

Checking plugin.audio.detektorfm
INFO: Created by sorax
INFO: Addon id matches folder name
INFO: Image icon exists
INFO: icon dimensions are fine 512x512
INFO: Image fanart exists
INFO: fanart dimensions are fine 1920x1080
WARN: Empty image tag found for screenshot

Checking plugin.audio.eco99music
INFO: Created by Nir Elbaz
INFO: Addon id matches folder name
INFO: Image icon exists
INFO: icon dimensions are fine 256x256
INFO: Image fanart exists
INFO: fanart dimensions are fine 1920x1080
INFO: Image screenshot exists
INFO: Image screenshot exists
INFO: Image screenshot exists

Checking plugin.audio.hearthis_at
INFO: Created by Eike Baran
INFO: Addon id matches folder name
INFO: Image icon exists
INFO: icon dimensions are fine 512x512
INFO: Image fanart exists
INFO: fanart dimensions are fine 1920x1080
INFO: Image screenshot exists
INFO: Image screenshot exists
INFO: Image screenshot exists
```

[Annotation](https://github.com/slgobinath/addon-check/blob/feature-new-architecture/kodi_addon_checker/plugins/console_reporter.py#L6) support is introduced for future implementations to automate config property definition which will be used to enable/disable reporters using `.test-config.json` or command line argument.

WDYT?

### DIFF
--- a/kodi_addon_checker/check_repo.py
+++ b/kodi_addon_checker/check_repo.py
@@ -8,7 +8,7 @@ from kodi_addon_checker.report import Report, INFORMATION, Record, ReportManager
 
 def check_repo(config, repo_path, parameters):
     repo_report = Report()
-    repo_report.log(Record(INFORMATION, "Checking repository %s" % repo_path))
+    repo_report.add(Record(INFORMATION, "Checking repository %s" % repo_path))
     print("Repo path " + repo_path)
     if len(parameters) == 0:
         toplevel_folders = sorted(next(os.walk(repo_path))[1])
@@ -19,10 +19,10 @@ def check_repo(config, repo_path, parameters):
 
     for addon_folder in toplevel_folders:
         if addon_folder[0] != '.':
-            repo_report.log(Record(INFORMATION, "Checking add-on %s" % addon_folder))
+            repo_report.add(Record(INFORMATION, "Checking add-on %s" % addon_folder))
             addon_path = os.path.join(repo_path, addon_folder)
             addon_report = check_addon.start(addon_path, config)
-            repo_report.log(addon_report)
+            repo_report.add(addon_report)
 
     ReportManager.report(repo_report)
 
--- a/kodi_addon_checker/check_repo.py
+++ b/kodi_addon_checker/check_repo.py
@@ -1,23 +1,12 @@
-import json
 import os
 import sys
 
 import kodi_addon_checker.check_addon as check_addon
 from kodi_addon_checker.common import colorPrint
-from kodi_addon_checker.plugins.console_reporter import ConsoleReporter
-from kodi_addon_checker.report import Report, INFORMATION, Record
+from kodi_addon_checker.report import Report, INFORMATION, Record, ReportManager
 
 
-def _read_config_for_version(repo_path):
-    config_path = os.path.join(repo_path, '.tests-config.json')
-    if os.path.isfile(config_path):
-        with open(config_path) as json_data:
-            return json.load(json_data)
-
-    return None
-
-
-def check_repo(repo_path, parameters):
+def check_repo(config, repo_path, parameters):
     repo_report = Report()
     repo_report.log(Record(INFORMATION, "Checking repository %s" % repo_path))
     print("Repo path " + repo_path)
@@ -28,8 +17,6 @@ def check_repo(repo_path, parameters):
 
     print("Toplevel folders " + str(toplevel_folders))
 
-    config = _read_config_for_version(repo_path)
-
     for addon_folder in toplevel_folders:
         if addon_folder[0] != '.':
             repo_report.log(Record(INFORMATION, "Checking add-on %s" % addon_folder))
@@ -37,9 +24,7 @@ def check_repo(repo_path, parameters):
             addon_report = check_addon.start(addon_path, config)
             repo_report.log(addon_report)
 
-    # Report using ConsoleReporter
-    reporter = ConsoleReporter()
-    reporter.report(repo_report)
+    ReportManager.report(repo_report)
 
     if repo_report.problem_count > 0:
         colorPrint("We found %s problems and %s warnings, please check the logfile." % (
--- a/kodi_addon_checker/check_repo.py
+++ b/kodi_addon_checker/check_repo.py
@@ -7,7 +7,7 @@ from kodi_addon_checker.report import Report, INFORMATION, Record, ReportManager
 
 
 def check_repo(config, repo_path, parameters):
-    repo_report = Report()
+    repo_report = Report(repo_path)
     repo_report.add(Record(INFORMATION, "Checking repository %s" % repo_path))
     print("Repo path " + repo_path)
     if len(parameters) == 0:
--- a/kodi_addon_checker/check_repo.py
+++ b/kodi_addon_checker/check_repo.py
@@ -5,7 +5,7 @@ import sys
 import kodi_addon_checker.check_addon as check_addon
 from kodi_addon_checker.common import colorPrint
 from kodi_addon_checker.plugins.console_reporter import ConsoleReporter
-from kodi_addon_checker.report import Report
+from kodi_addon_checker.report import Report, INFORMATION, Record
 
 
 def _read_config_for_version(repo_path):
@@ -18,7 +18,8 @@ def _read_config_for_version(repo_path):
 
 
 def check_repo(repo_path, parameters):
-    repo_report = Report(repo_path)
+    repo_report = Report()
+    repo_report.log(Record(INFORMATION, "Checking repository %s" % repo_path))
     print("Repo path " + repo_path)
     if len(parameters) == 0:
         toplevel_folders = sorted(next(os.walk(repo_path))[1])
@@ -31,6 +32,7 @@ def check_repo(repo_path, parameters):
 
     for addon_folder in toplevel_folders:
         if addon_folder[0] != '.':
+            repo_report.log(Record(INFORMATION, "Checking add-on %s" % addon_folder))
             addon_path = os.path.join(repo_path, addon_folder)
             addon_report = check_addon.start(addon_path, config)
             repo_report.log(addon_report)
@@ -39,13 +41,13 @@ def check_repo(repo_path, parameters):
     reporter = ConsoleReporter()
     reporter.report(repo_report)
 
-    if repo_report.problem > 0:
-        colorPrint("We found %s problem and %s warning, please check the logfile." % (
-            repo_report.problem, repo_report.warning), "31")
+    if repo_report.problem_count > 0:
+        colorPrint("We found %s problems and %s warnings, please check the logfile." % (
+            repo_report.problem_count, repo_report.warning_count), "31")
         sys.exit(1)
-    elif repo_report.warning > 0:
+    elif repo_report.warning_count > 0:
         # If we only found warning, don't mark the build as broken
-        colorPrint("We found %s problem and %s warning, please check the logfile." % (
-            repo_report.problem, repo_report.warning), "35")
+        colorPrint("We found %s problems and %s warnings, please check the logfile." % (
+            repo_report.problem_count, repo_report.warning_count), "35")
 
     print("Finished!")
--- a/kodi_addon_checker/check_repo.py
+++ b/kodi_addon_checker/check_repo.py
@@ -1,8 +1,11 @@
+import json
 import os
 import sys
-import json
+
 import kodi_addon_checker.check_addon as check_addon
 from kodi_addon_checker.common import colorPrint
+from kodi_addon_checker.plugins.console_reporter import ConsoleReporter
+from kodi_addon_checker.report import Report
 
 
 def _read_config_for_version(repo_path):
@@ -15,7 +18,7 @@ def _read_config_for_version(repo_path):
 
 
 def check_repo(repo_path, parameters):
-    error_counter = {"warnings": 0, "problems": 0}
+    repo_report = Report(repo_path)
     print("Repo path " + repo_path)
     if len(parameters) == 0:
         toplevel_folders = sorted(next(os.walk(repo_path))[1])
@@ -29,16 +32,20 @@ def check_repo(repo_path, parameters):
     for addon_folder in toplevel_folders:
         if addon_folder[0] != '.':
             addon_path = os.path.join(repo_path, addon_folder)
-            error_counter = check_addon.start(
-                error_counter, addon_path, config)
+            addon_report = check_addon.start(addon_path, config)
+            repo_report.log(addon_report)
 
-    if error_counter["problems"] > 0:
-        colorPrint("We found %s problems and %s warnings, please check the logfile." % (
-            error_counter["problems"], error_counter["warnings"]), "31")
+    # Report using ConsoleReporter
+    reporter = ConsoleReporter()
+    reporter.report(repo_report)
+
+    if repo_report.problem > 0:
+        colorPrint("We found %s problem and %s warning, please check the logfile." % (
+            repo_report.problem, repo_report.warning), "31")
         sys.exit(1)
-    elif error_counter["warnings"] > 0:
-        # If we only found warnings, don't mark the build as broken
-        colorPrint("We found %s problems and %s warnings, please check the logfile." % (
-            error_counter["problems"], error_counter["warnings"]), "35")
+    elif repo_report.warning > 0:
+        # If we only found warning, don't mark the build as broken
+        colorPrint("We found %s problem and %s warning, please check the logfile." % (
+            repo_report.problem, repo_report.warning), "35")
 
     print("Finished!")
--- a/kodi_addon_checker/check_repo.py
+++ b/kodi_addon_checker/check_repo.py
@@ -1,21 +1,18 @@
 import os
-import sys
 
 import kodi_addon_checker.check_addon as check_addon
-from kodi_addon_checker.common import colorPrint
-from kodi_addon_checker.report import Report, INFORMATION, Record, ReportManager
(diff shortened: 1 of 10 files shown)


enen92:
I like very much the idea, at least it decouples reporting from the checking step which is really usefull for creating integrations later. Two minor suggestions:

1) Try to avoid `log` as the method name. At first it made me think your reporters were extensions of the python logging module when in fact they are custom implementations. Use `report` for example.

2) Try to avoid importing reporters directly in checkers like your are doing right now. The checker should be agnostic to what reporters are available. That being said, and since decorators are executed when you import a module it would make more sense to define the import logic in the `__init__.py` file of the reporter module using the `__all___` variable for example. By doing this (or a similar logic) you are creating a sort of "reporterManager" which holds references to all available reporters. The manager would also be responsible for ignoring/selecting reporters if they were somehow defined in a configuration file.

slgobinath:
Hi @enen92,
> Try to avoid log as the method name

Sure

> Try to avoid importing reporters directly in checkers like your are doing right now

You have read my mind :-)
That is the plan. This PR is to get an approval. I will implement ReportManager, ConfigManager and any other required modules to automate importing reporters and enable/disable them depending on configuration. The decorator added in this PR is for that purpose but not yet used for automatic injection.
I will work on that.

kodi_addon_checker/report.py
```diff
@@ -0,0 +1,80 @@
...
+        self.problem += report.problem
+        self.warning += report.warning
+        self.information += report.information
+
+    def __iter__(self):
+        return iter(self.reports)
+
+
+class Record(Report):
+    def __init__(self, artifact_name, message, start_line=-1, end_line=-1, start_char_position=-1,
```
Razzeee: I'm not convinced by the naming of `artifact_name` here. Would be clearer as log level or similar.

kodi_addon_checker/report.py
```diff
@@ -0,0 +1,80 @@
...
+                 end_char_position=-1):
+        super().__init__(artifact_name)
+        self.message = message
+        self.start_line = start_line
+        self.end_line = end_line
+        self.start_char_position = start_char_position
+        self.end_char_position = end_char_position
+
+        if PROBLEM == artifact_name:
```
Razzeee: I don't think we should have fields for problem counts (etc) directly on the record type. Move this logic to the report log method. And get rid of these fields here.

Razzeee:
Very very good direction to move things in. So definatly welcome. Keep up the great work. 🎉

kodi_addon_checker/plugins/console_reporter.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def report(self, report: Report):
+        if type(report) is Record:
+            if report.information:
+                colorPrint(report, "34")
+            elif report.warning:
+                colorPrint(report, "35")
+            elif report.problem:
+                colorPrint(report, "31")
+            else:
```
Razzeee: I don't think we need an else?

kodi_addon_checker/plugins/console_reporter.py
```diff
@@ -0,0 +1,22 @@
...
+                print(report)
+        else:
+            print("\nChecking %s" % report.artifact_name)
```
Razzeee: You can just use the info log level for this and drop this else

kodi_addon_checker/check_repo.py
```diff
@@ -29,16 +32,20 @@ def check_repo(repo_path, parameters):
...
+
+    # Report using ConsoleReporter
+    reporter = ConsoleReporter()
+    reporter.report(repo_report)
 
-    if error_counter["problems"] > 0:
-        colorPrint("We found %s problems and %s warnings, please check the logfile." % (
-            error_counter["problems"], error_counter["warnings"]), "31")
+    if repo_report.problem > 0:
+        colorPrint("We found %s problem and %s warning, please check the logfile." % (
```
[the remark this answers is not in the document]
slgobinath: You are right. I will fix it :-)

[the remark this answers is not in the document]
slgobinath: Fixed

kodi_addon_checker/check_repo.py
```diff
@@ -29,16 +32,20 @@ def check_repo(repo_path, parameters):
...
+            repo_report.problem, repo_report.warning), "31")
         sys.exit(1)
-    elif error_counter["warnings"] > 0:
-        # If we only found warnings, don't mark the build as broken
-        colorPrint("We found %s problems and %s warnings, please check the logfile." % (
-            error_counter["problems"], error_counter["warnings"]), "35")
+    elif repo_report.warning > 0:
+        # If we only found warning, don't mark the build as broken
+        colorPrint("We found %s problem and %s warning, please check the logfile." % (
```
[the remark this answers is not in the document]
slgobinath: Fixed

kodi_addon_checker/plugins/console_reporter.py
slgobinath: Fixed

kodi_addon_checker/plugins/console_reporter.py
slgobinath: Fixed by removing `print` and adding info logs before checking repos and add-ons. However, `else` is required for recursion.

kodi_addon_checker/report.py
```diff
@@ -0,0 +1,80 @@
...
+WARNING = "WARN"
+INFORMATION = "INFO"
+
+KODI_REPORTERS = {}
+
+
+class Report(object):
+    def __init__(self, artifact_name):
+        self.artifact_name = artifact_name
+        self.problem = 0
```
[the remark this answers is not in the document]
slgobinath: Fixed

kodi_addon_checker/report.py
slgobinath: Fixed by renaming to `log_level`

kodi_addon_checker/report.py
slgobinath: Fixed

slgobinath:
Hi @enen92,
I tried renaming the function name `log` to `report` but then it looks odd when we call it using `Report` object which is named `report`
For example:
```
report.report(Record(INFORMATION, "Created by %s" % addon.attrib.get("provider-name")))
```
Shall we find a different name or leave `log` for the moment? WDYT?

Razzeee:
Add might work, as we basically add the record / report item and do some processing

slgobinath:
Renamed `log` to `add`.
With the commit https://github.com/xbmc/addon-check/pull/64/commits/cd194c4e687c591ab911e8e33e64401085966919, reporters are injected and managed automatically. A new configuration `reporter` is added to both `.test-config.json` and command line arguments which are used to enable reporters.

By default, Console reporter is enabled (`@reporter(name="console", enabled=True)`). If you provide `"reporter": []` in `.test-config.json` as below, it will disable all reporters.
```
{
    "check_legacy_language_path": true,
    "reporter": []
}
```

The user can override that option in the command line by providing `--reporter console`. The reporter name `console` is determined by the decorator `@reporter(name='console'...)`. Therefore, if you add a new reporter, you just need to define the name and whether it is enabled by default.

```
$ kodi-addon-checker --help
usage: kodi-addon-checker [-h] [--version] [--reporter {console}]
                          [add-on [add-on ...]]

Checks Kodi repo for best practices and creates problem and warning reports.
If optional add-on directories are provided, check only those add-ons.
Otherwise, scan current repository and check all add-ons in the current
directory.

positional arguments:
  add-on                optional add-on directories

optional arguments:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  --reporter {console}  enable a reporter with the given id(s). You can put
                        this option multiple time

```
If there are more than one reporters (say `console` and `github`) user can enable both of them using the following command:
```
kodi-addon-checker --reporter console --reporter github
```
I feel it is enough for a PR. I will improve the file checker architecture in another PR. Please let me know if there is anything to fix.

kodi_addon_checker/config.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @classmethod
+    def register(cls, config, description, default_value, action):
+        cls.configurations[config] = [description, default_value, action]
+
+    @classmethod
+    def fill_cmd_args(cls, parser: ArgumentParser):
+        # Add --reporter
+        parser.add_argument("--reporter", action="append", choices=list(ReportManager.reporters.keys()),
+                            help="""enable a reporter with the given name.
```
Razzeee: We probably can't print the available reports here right? Can we add them manually at least then?

slgobinath: We do in the `--help` message. Please check:
```
  --reporter {console}  enable a reporter with the given name. You can use
                        this option multiple times to enable more than one
                        reporters
```
where whatever comes in between `{...}` are the available options.

kodi_addon_checker/report.py
```diff
@@ -0,0 +1,105 @@
...
+    def __iter__(self):
+        return iter(self.reports)
+
+
+class Record(Report):
+    def __init__(self, log_level, message, start_line=-1, end_line=-1, start_char_position=-1,
+                 end_char_position=-1):
+        self.log_level = log_level
+        self.message = message
+        self.start_line = start_line
```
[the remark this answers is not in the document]
Razzeee: I'm confused by this. I might not understand the code here. But your link seems like you only hold the reference to the addon folder/path. So how do we know on a Record for which file it was, especially if it has a linenumber etc.

Razzeee:
Well maybe it's not needed / not a good fit to derrive Record from Report?

slgobinath:
@Razzeee, please note that the inheritance has been removed in https://github.com/xbmc/addon-check/pull/64/commits/d1235ed1913690790065b68555696beb47b8376d

Razzeee:
I'll try to review later today, thank you!

kodi_addon_checker/report.py
```diff
@@ -0,0 +1,106 @@
...
+            self.problem_count += report.problem_count
+            self.warning_count += report.warning_count
+            self.information_count += report.information_count
+
+    def __iter__(self):
+        return iter(self.reports)
+
+
+class Record(object):
+    def __init__(self, log_level, message, start_line=-1, end_line=-1, start_char_position=-1,
```
Razzeee: Not sure if that's the python way to do things, but please give these classes their own files :)

kodi_addon_checker/check_repo.py
```diff
@@ -1,21 +1,14 @@
...
-
-    return None
-
-
-def check_repo(repo_path, parameters):
-    error_counter = {"warnings": 0, "problems": 0}
+def check_repo(config, repo_path, parameters):
+    repo_report = Report(repo_path)
+    repo_report.add(Record(INFORMATION, "Checking repository %s" % repo_path))
     print("Repo path " + repo_path)
```
Razzeee: I think we can remove this line, as it's logged above

kodi_addon_checker/check_repo.py
```diff
@@ -24,21 +17,22 @@ def check_repo(repo_path, parameters):
 
     print("Toplevel folders " + str(toplevel_folders))
```
Razzeee: We should probably not use print at all anymore, right? So let's use the reporter here also.

kodi_addon_checker/check_repo.py
```diff
@@ -24,21 +17,22 @@ def check_repo(repo_path, parameters):
...
         sys.exit(1)
-    elif error_counter["warnings"] > 0:
-        # If we only found warnings, don't mark the build as broken
+    elif repo_report.warning_count > 0:
+        # If we only found warning, don't mark the build as broken
         colorPrint("We found %s problems and %s warnings, please check the logfile." % (
-            error_counter["problems"], error_counter["warnings"]), "35")
+            repo_report.problem_count, repo_report.warning_count), "35")
 
     print("Finished!")
```
Razzeee: Another use of print

kodi_addon_checker/common.py
```diff
@@ -1,14 +1,11 @@
...
 
 
-def check_config(config, value):
-    if config is None:
-        return False
-    elif value in config and config[value]:
-        return True
-    else:
-        return False
+def colorPrint(string, color):
```
Razzeee: I think we should move this method to `console_reporter.py` as it's console specific.

kodi_addon_checker/check_repo.py
slgobinath: Yeah.
Shall we completely remove this part because we also have `INFO: checking plugin.foo.bar`?

kodi_addon_checker/check_repo.py
```diff
@@ -24,21 +17,22 @@ def check_repo(repo_path, parameters):
...
-            error_counter = check_addon.start(
-                error_counter, addon_path, config)
+            addon_report = check_addon.start(addon_path, config)
+            repo_report.add(addon_report)
+
+    ReportManager.report(repo_report)
 
-    if error_counter["problems"] > 0:
+    if repo_report.problem_count > 0:
         colorPrint("We found %s problems and %s warnings, please check the logfile." % (
```
Razzeee: Is this correct? Shouldn't this also use the reporter and just create a record?

kodi_addon_checker/check_repo.py
```diff
@@ -24,21 +17,22 @@ def check_repo(repo_path, parameters):
...
-            error_counter["problems"], error_counter["warnings"]), "31")
+            repo_report.problem_count, repo_report.warning_count), "31")
         sys.exit(1)
-    elif error_counter["warnings"] > 0:
-        # If we only found warnings, don't mark the build as broken
+    elif repo_report.warning_count > 0:
+        # If we only found warning, don't mark the build as broken
         colorPrint("We found %s problems and %s warnings, please check the logfile." % (
```
Razzeee: See comment above

Razzeee:
@slgobinath I just added some suggestions.
Do you think it's better without the inheritance? Just curious, what you think.

kodi_addon_checker/check_repo.py
Razzeee: I would like to keep this, this will print all folders it gathered, if we run it on a whole repo.

slgobinath: I mean even no need for Reporter at this place.

kodi_addon_checker/common.py
slgobinath: I thought about it but later didn't do that because it will create a dependency between addon-checker and `ConsoleReporter`. In my opinion, ConsoleReporter and whatever we have in `plugins` package should not be accessed directly from add-on checker.
If we move `colorPrint` to console reporter, I would recommend using normal `print` everywhere else.
WDYT?

slgobinath: Sorry.
After seeing your other comments, now it makes sense. I am removing all `colorPrint` calls.

kodi_addon_checker/check_repo.py
Razzeee: I don't think I get what your saying :/

But can you agree that it's useful information?

kodi_addon_checker/report.py
slgobinath: Fixed

kodi_addon_checker/check_repo.py
slgobinath: Fixed

kodi_addon_checker/check_repo.py
slgobinath: Removed the line

kodi_addon_checker/check_repo.py
slgobinath: Fixed

kodi_addon_checker/check_repo.py
slgobinath: Fixed

kodi_addon_checker/check_repo.py
slgobinath: Fixed

slgobinath:
@Razzeee please note that I have fixed all your suggestions in https://github.com/slgobinath/addon-check/commit/8fb6fc01feef886007c85303cc6f1452ad0d9342

> Do you think it's better without the inheritance?

I am from strong OOP and strict type background (Java person) so I was too concern about exact relationships. My design was "Record IS A Report" and "Report HAS A Report". However, I find removing inheritance makes the code more readable and does not hurt in Python world.

Razzeee:
Thank you very much for this 🎉 
Really like the comments you added in the last commit.

PR #65, opened 2018-04-09
Add launch.json config for vscode & always report when finished

Razzeee:
@slgobinath Will you implement a repo parameter or should I look at it?

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,23 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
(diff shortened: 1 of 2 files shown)


slgobinath:
Hi @Razzeee,
I think we can do it without repo parameter. My plan is looking for `addon.xml` in passed directories. If `addon.xml` is available treat it as add-on folder otherwise repo. If that sounds good, PR is on the way :)

Razzeee:
yes, sounds awesome. Be sure to check for each path you get, othen then that pretty straight forward.

PR #66, opened 2018-04-09
Support multiple repos and add-ons through cmd args

slgobinath:
@Razzeee 
This PR supports multiple add-ons and/or repositories to be passed as command line argument. If none is passed, treat current directory as a repo.

For repositories their respective `.tests-config.json` is used. For add-ons, if there is a `tests-config.json` in the current directory, that will be used.

Tested using two repos (krypton and jarvis) using different `.tests-config.json` settings.

Please note that the command line argument is renamed from `add-on` to `dir`. If you have a better name, please let me know.

Thanks

### DIFF
--- a/kodi_addon_checker/__main__.py
+++ b/kodi_addon_checker/__main__.py
@@ -5,7 +5,7 @@ from kodi_addon_checker import check_addon
 from kodi_addon_checker.check_repo import check_repo
 from kodi_addon_checker.common import load_plugins
 from kodi_addon_checker.config import ConfigManager, Config
-from kodi_addon_checker.record import Record, INFORMATION, PROBLEM, WARNING
+from kodi_addon_checker.record import Record, PROBLEM, WARNING
 from kodi_addon_checker.report import Report
 from kodi_addon_checker.reporter import ReportManager
 
@@ -28,6 +28,22 @@ def dir_type(dir_path):
     return os.path.abspath(dir_path)
 
 
+def check_artifact(artifact_path, args):
+    """
+    Check given artifact and return its report. The artifact can be either an add-on or a repository.
+    :param artifact_path: the path of add-on or repo
+    :param args: argparse object
+    :return: report
+    """
+    artifact_path = os.path.abspath(artifact_path)
+    config = Config(artifact_path, args)
+    ConfigManager.process_config(config)
+    if os.path.isfile(os.path.join(artifact_path, "addon.xml")):
+        return check_addon.start(artifact_path, config)
+    else:
+        return check_repo(artifact_path, config)
+
+
 def main():
     """The entry point to kodi-addon-checker
     """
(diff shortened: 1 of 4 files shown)


kodi_addon_checker/__main__.py
```diff
@@ -39,15 +40,47 @@ def main():
...
-                        help="optional add-on directories")
+    parser.add_argument("dir", type=dir_type, nargs="*", help="optional add-on or repo directories")
     ConfigManager.fill_cmd_args(parser)
     args = parser.parse_args()
 
-    repo_path = os.path.abspath(os.getcwd())
-    config = Config(repo_path, args)
-    ConfigManager.process_config(config)
-    check_repo(config, repo_path, args.add_on)
+    current_dir = os.path.abspath(os.getcwd())
```
Razzeee: If we pass the directory to the checker, this folder might be completly unrelated. Therefore I think we should never use this in that case.

kodi_addon_checker/__main__.py
```diff
@@ -39,15 +40,47 @@ def main():
...
+    if args.dir:
+        report = Report(current_dir)
+        for directory in args.dir:
+            if os.path.isfile(os.path.join(directory, "addon.xml")):
+                report.add(Record(INFORMATION, "Checking add-on %s" % directory))
+                # For add-ons try to load .tests-config.json from current directory
+                config = Config(current_dir, args)
```
Razzeee: Go for `os.path.abspath(directory)` here. That way developers can just have a local gitignored copy of the .test-config.json they want to use int the addons root dir.

kodi_addon_checker/__main__.py
```diff
@@ -39,15 +40,47 @@ def main():
...
+                addon_report = check_addon.start(os.path.abspath(directory), config)
+                report.add(addon_report)
+            else:
+                repo_path = os.path.abspath(directory)
```
Razzeee: Can we print something for this case too? Just to communicate that we're using repo handling?

kodi_addon_checker/__main__.py
```diff
@@ -39,15 +40,47 @@ def main():
...
+                config = Config(repo_path, args)
+                repo_report = check_repo(repo_path, config)
+                report.add(repo_report)
+    else:
+        if os.path.isfile(os.path.join(current_dir, "addon.xml")):
+            # Current directory is an add-on
+            report = Report(current_dir)
+            report.add(Record(INFORMATION, "Checking add-on %s" % current_dir))
+            config = Config(None, args)
+            addon_report = check_addon.start(os.path.abspath(current_dir), config)
```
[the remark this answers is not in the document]
slgobinath: I have moved `Checking ...` INFO to relevant checkers. Now check_repo adds `Checking repository ...` and check_addon.start adds `Checking add-on ...` which avoids code duplication and improve cohesion.

Rechi:
@slgobinath please remove the merge commit.
If you want the latest changes of the base branch in your feature branch rebase on it instead of merge it back.

slgobinath:
@Rechi and @Razzeee 
Please note that I am closing this PR in favor of https://github.com/xbmc/addon-check/pull/67 with clean commits history.

PR #69, opened 2018-04-12
Add check for complexity of entry point files

### LINKED ISSUE #68: Check complexity of addon entrypoints
It is a general good practice to keep the entrypoints (the python files refered in addon.xml) to a minimum since they won't get compiled to bytecode. Usually the entrypoint should be something as simpler as:

```python
import resources.lib.plugin

plugin.run()
```

While the main plugin code is kept in a module.
The `addon-check-tool` could somehow check the complexity of the entrypoint and might suggest the author to improve the addon performance by pointing to an example. A few possibilities:

1) Just check the number of uncommented lines in the entrypoint file and decide if it is lower than, lets say, 10 lines?

2) Check the number of imports on the entrypoint? This probably means the main plugin logic is executed there or, at least, there is room for improvement

3) Something more complex by exploiting cyclomatic complexity? Might be too hard to accomplish.

4) Infer from the number of python files in the addon that are not the entrypoint. This can lead to a lot of false positives though.

Option 1) seems the easiest one to implement if reforced by a few more checks

mzfr:
Fixes #68 

I have taken an example `.tests-config.json`

```json
{
    "check_legacy_language_path": false,
    "comment_on_pull": false,
    "check_dependencies": true,
    "check_complex_files": true,
    "max_line_count": 50,
}
```
Here function name is `check_complex_files` that gets all the extensions from the addon.xml and performs the check and `max_line_count` stores maximum number of lines allowed that will not raise any 
warning.

@enen92 Please let me know if any changes are required.

### DIFF
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -3,6 +3,7 @@ import os
 import pathlib
 import re
 import xml.etree.ElementTree
+from radon.raw import analyze
 
 from PIL import Image
 
@@ -79,7 +80,6 @@ def start(addon_path, config=None):
             _check_artwork(addon_report, addon_path, addon_xml, file_index)
 
             max_entrypoint_line_count = config.configs.get("max_entrypoint_line_count", 10)
-            print(max_entrypoint_line_count)
             _check_complex_addon_entrypoint(addon_report, addon_path, max_entrypoint_line_count)
 
             if config.is_enabled("check_license_file_exists"):
@@ -329,20 +329,7 @@ def _check_complex_addon_entrypoint(report: Report, addon_path, max_entrypoint_l
 
 
 def number_of_lines(filepath):
-    lineno = 0
-    skip = False
+    with open(filepath, 'r') as file:
+        data = file.read()
 
-    with open(filepath) as fp:
-
-        for line in fp:
-            line = line.strip()
-
-            if line:
-                if line.startswith('#'):
-                    continue
-                if line.startswith('"""'):
-                    skip = not skip
-                    continue
-                if not skip:
-                    lineno += 1
-    return lineno
+    return(analyze(data).lloc)
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -332,4 +332,4 @@ def number_of_lines(filepath):
     with open(filepath, 'r') as file:
         data = file.read()
 
-    return(analyze(data).lloc)
+    return (analyze(data).lloc)
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -322,8 +322,12 @@ def _check_complex_addon_entrypoint(report: Report, addon_path, max_entrypoint_l
             if not os.path.isdir(filepath):
 
                 if os.path.exists(filepath):
-                    if number_of_lines(filepath) > max_entrypoint_line_count:
-                        report.add(Record(WARNING, "Entry point for the addon is complex. please check %s" % library))
+                    lineno = number_of_lines(filepath)
+                    if lineno >= max_entrypoint_line_count:
+                        report.add(Record(WARNING,
+                                          "Complex entry point. Check: %s | Counted lines: %d | Lines allowed: %d"
+                                          % (library, lineno, max_entrypoint_line_count)))
+
                 else:
                     report.add(Record(PROBLEM, "%s Entry point does not exists" % library))
 
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -78,9 +78,9 @@ def start(addon_path, config=None):
 
             _check_artwork(addon_report, addon_path, addon_xml, file_index)
 
-            if config.is_enabled("check_complex_files"):
-                max_line_count = config.configs.get("max_line_count", 80)
-                _check_complex_files(addon_report, addon_path, max_line_count)
+            max_entrypoint_line_count = config.configs.get("max_entrypoint_line_count", 10)
+            print(max_entrypoint_line_count)
+            _check_complex_addon_entrypoint(addon_report, addon_path, max_entrypoint_line_count)
 
             if config.is_enabled("check_license_file_exists"):
                 # check if license file is existing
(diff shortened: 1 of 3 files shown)


kodi_addon_checker/check_addon.py
```diff
@@ -78,6 +78,10 @@ def start(addon_path, config=None):
 
             _check_artwork(addon_report, addon_path, addon_xml, file_index)
 
+            if config.is_enabled("check_complex_files"):
+                max_line_count = config.configs.get("max_line_count", 80)
```
mzfr: @enen92 here i have taken a default value for `max_line_count`so if it is not present in any config file then it will use a default value.
Should I leave it to 80 ?

kodi_addon_checker/check_addon.py
```diff
@@ -302,3 +306,19 @@ def _check_file_whitelist(report: Report, file_index, addon_path):
...
+
+
+def _check_complex_files(report: Report, addon_path, max_line_count):
+
+    addon_xml_path = os.path.join(addon_path, "addon.xml")
+    tree = xml.etree.ElementTree.parse(addon_xml_path).getroot()
+    complex_files = []
+
+    for i in tree.findall("extension"):
+        library = i.get("library")
```
enen92: This won't work for context menu items library is not an attribute of the extension element but an attribute for the item child element

kodi_addon_checker/check_addon.py
```diff
@@ -302,3 +306,19 @@ def _check_file_whitelist(report: Report, file_index, addon_path):
...
+        if library:
+            filepath = os.path.join(addon_path, library)
+            num_line = sum(1 for line in open(filepath))
```
enen92: please use `with` python statement so that `open(filepath)` is automatically closed:

```python
with open(filepath) as fp:
    num_line = sum(1 for line in fp)
```

kodi_addon_checker/check_addon.py
```diff
@@ -302,3 +306,19 @@ def _check_file_whitelist(report: Report, file_index, addon_path):
...
+def _check_complex_files(report: Report, addon_path, max_line_count):
+
+    addon_xml_path = os.path.join(addon_path, "addon.xml")
+    tree = xml.etree.ElementTree.parse(addon_xml_path).getroot()
+    complex_files = []
+
+    for i in tree.findall("extension"):
+        library = i.get("library")
+        if library:
+            filepath = os.path.join(addon_path, library)
```
enen92: please check if filepath exists with `os.path.exists` and if not produce an ERROR message for an invalid/unexistent entrypoint

kodi_addon_checker/check_addon.py
enen92: I'd keep it much lower than that as a default like 10. However there is an issue. 
You're blindingly counting the lines of code and you basically neglect the presence of comments. Some addon devs keep and header at the top with a copyright message that will increase the number of lines in the file but won't get executed by the interpreter

enen92:
I'd change `check_complex_files` to `check_complex_addon_entrypoints`.  `max_line_count` to `max_entrypoint_line_count`

kodi_addon_checker/check_addon.py
enen92: See:
https://github.com/enen92/screensaver.atv4/blob/master/entrypointscript.py

The tool would report 26 lines of code when in fact the entrypoint file is only composed of 3

mzfr: ok so i'll make another function for checking the line which will ignore docstring and comments

kodi_addon_checker/check_addon.py
mzfr: For this we'll have to get its children item so i'll have to make another iterator for it right ? because for other this thing works fine.

kodi_addon_checker/check_addon.py
```diff
@@ -302,3 +306,43 @@ def _check_file_whitelist(report: Report, file_index, addon_path):
...
+            if not os.path.isdir(filepath):
+
+                if os.path.exists(filepath):
+                    if number_of_lines(filepath) > max_entrypoint_line_count:
+                        report.add(Record(WARNING, "Entry point for the addon is complex. please check %s" % library))
+                else:
+                    report.add(Record(PROBLEM, "%s Entry point does not exists" % library))
+
+
+def number_of_lines(filepath):
```
mzfr: @enen92 This is very crude way of doing this, that does not involve ast module. But if we want to go that way, I can work on that too!

kodi_addon_checker/check_addon.py
```diff
@@ -78,6 +78,10 @@ def start(addon_path, config=None):
 
             _check_artwork(addon_report, addon_path, addon_xml, file_index)
 
+            max_entrypoint_line_count = config.configs.get("max_entrypoint_line_count", 10)
+            print(max_entrypoint_line_count)
```
Rechi: remove

Rechi:
The line counter doesn't work correctly. If I add `""""""` at the beginning of my entry point file it counts zero lines.

mzfr:
@Rechi using ast module for counting the number of line can be a bit of complex but we can use Radon to do the work.
Radon have an option for displaying raw data about the file like lloc, sloc, total lines. This will be a very easy way to do it.
All we have to do it is read the content of the whole file and then pass it to radon to do its work.
The code would look like 
```python
from radon.raw import analyze
def number_of_lines(filename):
    with open(filename, 'r') as myfile:
         data=myfile.read()

    return(analyze(data).lloc)
```

This would return number of logical line of code(lloc).

For the [file](https://github.com/enen92/screensaver.atv4/blob/master/entrypointscript.py)mentioned by @enen92 we will get `3` as an output.

Please let me know what do you think about it ?

Razzeee:
Well was radon in your pitch for gsoc? Otherwise, just do the thing that's easiest and works reliably.

mzfr:
Yes, there I suggested to use it for code complexity detection. Here we'll use the LOC metric. Radon is the simplest and most reliable way I can think of.

kodi_addon_checker/check_addon.py
```diff
@@ -329,20 +329,7 @@ def _check_complex_addon_entrypoint(report: Report, addon_path, max_entrypoint_l
...
-                if line.startswith('"""'):
-                    skip = not skip
-                    continue
-                if not skip:
-                    lineno += 1
-    return lineno
+    with open(filepath, 'r') as file:
+        data = file.read()
+
+    return(analyze(data).lloc)
```
Razzeee: add a space please :)

mzfr: Done

Razzeee:
Radon needs to be installed in the travis-ci examples

kodi_addon_checker/check_addon.py
```diff
@@ -302,3 +306,30 @@ def _check_file_whitelist(report: Report, file_index, addon_path):
...
+        library = i.get("library")
+
+        if library:
+            filepath = os.path.join(addon_path, library)
+
+            if not os.path.isdir(filepath):
+
+                if os.path.exists(filepath):
+                    if number_of_lines(filepath) > max_entrypoint_line_count:
+                        report.add(Record(WARNING, "Entry point for the addon is complex. please check %s" % library))
```
Razzeee: This error message needs work. Would be helpful to mention how many lines we counted vs how many are allowed.
Maybe we should start to document warnings/problems in our wiki and link to those documentations in the errors?

kodi_addon_checker/check_addon.py
```diff
@@ -302,3 +306,30 @@ def _check_file_whitelist(report: Report, file_index, addon_path):
...
+    for i in tree.findall("extension"):
+        library = i.get("library")
+
+        if library:
+            filepath = os.path.join(addon_path, library)
+
+            if not os.path.isdir(filepath):
+
+                if os.path.exists(filepath):
+                    if number_of_lines(filepath) > max_entrypoint_line_count:
```
Razzeee: I think we want >=

mzfr:
@Razzeee I have made the changes suggested by you. Please review them.

PR #72, opened 2018-05-22
Test addon 

mzfr:
This is a test addon that we'll use for working of our tests.

It contains
* A correct size icon(icon2.png)
* An incorrect size icon(icon.png)
* An incorrect size fanart(fanart.jpg)
* A complex entry point file(main.py)

The main.py file have all the code of addon-check tool. This was done to increase the size the entry point file.


Razzeee:
Which also creates the question why we don't report an error if you declare two icons. As you can't have two icons at once.

Razzeee:
Looks fine, but two things:

- Where is the fanart from? Do we need to ask the original author and who is it?
- Can you look into bundling for pypi and ignoring this new sub folder for such builds?

mzfr:
yes I agree we should do that.
We can just check whether there are more than one icon/fanarts tags present in the addon.xml

I took the fanart from [here](https://forum.kodi.tv/archive/index.php?thread-209706-5.html)

mzfr:
@Razzeee Can I use fanart from one of your plugins i.e [plugin.video.rocketbeans](https://github.com/Razzeee/plugin.video.rocketbeans/blob/master/fanart.jpg)

mzfr:
@Razzeee I have uploaded a new fanart :slightly_smiling_face:

Razzeee:
We might want to go back and add other python code for the test addon, makes it quiet hard to search for stuff. You will always find two occurrences at least now.

zag2me:
Sorry for the random drive by, but shouldn't icon and fanart be in the root to provide backward compatibility with older kodi versions?

Razzeee:
No, if you don't want to be backward compatible. I usually don't do that, as I only move forward.

And in this case it's even better, as we need cases to test against.

mzfr:
Also it wouldn't matter if we keep it in root folder or not because our checks are like check in resource folder if not present there then look in root path so we don't have to worry about where we keep them.

Razzeee:
We still could test both cases working, but that would require another addon to test against

PR #74, opened 2018-05-26
Check for multiple artwork(icon/fanart)

mzfr:
Add check for multiple Icons/fanart.
After noticing multiple icons in [test addon](https://github.com/xbmc/addon-check/tree/master/script.test) @Razzeee pointed out that we should have a check for multiple artworks


Rechi:
This should be checked by XML schema validation.

mzfr:
@Rechi ok I did not had a detailed discussion on this one with @Razzeee.
We can do this with schemas validation but my PR#35 (schema validation) is stuck because I am not sure in what way I am supposed to [complete schemas](https://github.com/xbmc/addon-check/pull/35#issuecomment-380799149) and also haven't got any reply from @tamland.

PR #77, opened 2018-05-29
Use libversion instead of LooseVersion

Razzeee:
Travis is failing at the moment when we try to run it against our repos. https://travis-ci.org/xbmc/repo-plugins/builds/385427172
And LooseVersion seems to be an internal interface in a deprecated module.

So this is a try to get around that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
 
 # command to install dependencies
 install:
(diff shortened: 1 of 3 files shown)


Razzeee:
It fails due to required_version being None.
So we moght be able to guard against that as a workaround instea.

PR #80, opened 2018-05-30
Catch error when parsing py files with radon

Razzeee:
This was causing some problems when we tried to run on a whole repo, as we have addons that have .py files defined as UtF-8-BOM.

@Rechi  @mzfr


mzfr:
@Razzeee Thanks for doing it. How did you figured it out that it was encoding issue ?

Razzeee:
Added some prints and had to step through the code with the debugger. Then it was just a guess, when I saw nothing wrong with the syntax. The radon docs might have also tipped me off on that.

PR #81, opened 2018-05-31
Create a logger for debugging

mzfr:
This will help us to debug without messing up the Travis-CI logs. It takes an argument `--log_file` which will be the `path of the log file`. If this argument is not given then by default log file will be created in the current working directory. The name of the log file is `kodi-addon-checker.log`.

The output in the log file will look like:
```
2018-05-31 14:16:38 INFO:kodi_addon_checker.check_addon:start: Hello logging
```
**Note**
Right now there is no logging set in any of the file. But we'll be able to use this logger in future.

### DIFF
--- a/kodi_addon_checker/__main__.py
+++ b/kodi_addon_checker/__main__.py
@@ -1,5 +1,7 @@
 import argparse
 import os
+import logging
+from kodi_addon_checker import logger
 from kodi_addon_checker import check_addon
 from kodi_addon_checker.check_repo import check_repo
 from kodi_addon_checker.common import load_plugins
@@ -64,9 +66,11 @@ def main():
     parser.add_argument("dir", type=dir_type, nargs="*", help="optional add-on or repo directories")
     parser.add_argument("--branch", choices=choice, required=True,
                         help="Target branch name where the checker will resolve dependencies")
-
     ConfigManager.fill_cmd_args(parser)
     args = parser.parse_args()
+
+    log_file_name = os.path.join(os.getcwd(), "kodi-addon-checker.log")
+    logger.Logger.create_logger(log_file_name, __package__)
 
     if args.dir:
         # Following report is a wrapper for all sub reports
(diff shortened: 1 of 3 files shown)


kodi_addon_checker/__main__.py
```diff
@@ -64,10 +66,16 @@ def main():
     parser.add_argument("dir", type=dir_type, nargs="*", help="optional add-on or repo directories")
     parser.add_argument("--branch", choices=choice, required=True,
                         help="Target branch name where the checker will resolve dependencies")
-
+    parser.add_argument("--log_file", help="path to log file")
```
Razzeee: let's do a simple `--logfile` instead

Razzeee: Or well you can drop this completely.
My original idea was to switch on this and if we get this also print to the console. But this is good enough, if you ask me.

Razzeee: Correct, I don't think we have any usecase for that

mzfr: Fixed

kodi_addon_checker/check_addon.py
```diff
@@ -64,6 +65,7 @@ def start(addon_path, repo_addons, config=None):
     addon_id = os.path.basename(os.path.normpath(addon_path))
     addon_report = Report(addon_id)
     addon_report.add(Record(INFORMATION, "Checking add-on %s" % addon_id))
+    logger = logging.getLogger(__name__)
```
Razzeee: Setting this up should be the first thing we do. We can also think about miving it to a constructor, if that's possible.

mzfr: Do mean right after this `def start(addon_path, repo_addons, config=None):
` line ?

Razzeee: yes

Razzeee:
Another thing is that this only appends to the logfile. It would be good to clear the file whenever we start writing the file. So that we always only have the logfile of the latest run.

mzfr:
@Razzeee I have made the suggested changes

PR #84, opened 2018-06-04
Tests for kodi-addon-checker

mzfr:
This is an integration test for the kodi-addon-checker tool.

I have made the import for the `__main__.py`  in `__init.py` file because it is consider a bad practice to do something like: `from __main__ import foo`.

### DIFF
--- a/tests/test_check_addon.py
+++ b/tests/test_check_addon.py
@@ -1,4 +1,5 @@
 import unittest
+from kodi_addon_checker import _all_repo_addons
 from kodi_addon_checker.check_addon import start
 from kodi_addon_checker.record import Record
 from kodi_addon_checker.config import Config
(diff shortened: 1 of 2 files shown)


tests/test_check_addon.py
```diff
@@ -0,0 +1,23 @@
...
+from kodi_addon_checker import _all_repo_addons
+from kodi_addon_checker.check_addon import start
+from kodi_addon_checker.record import Record
+from kodi_addon_checker.config import Config
+
+
+class TestCheckAddon(unittest.TestCase):
+    """Integration tests for Start function present in check_addon.py"""
+
+    def setUp(self):
```
Razzeee: Just `setup`, as far as I know it's one word

tests/test_check_addon.py
```diff
@@ -0,0 +1,23 @@
...
+                          "Image icon", "Icon dimensions are fine", "INFO: Image fanart",
+                          "WARN: Complex entry point", "WARN: We found", "please check the logfile"]
+        self.config = Config(self.path)
+        self.branch_name = "krypton"
+        self.all_repo_addons = _all_repo_addons()
+
+    def test_start(self):
+        result = start(self.path, self.branch_name, self.all_repo_addons, self.config)
+        values = [Record.__str__(r) for r in result]
+        self.assertTrue(any(any(x in z for z in values) for x in self.whitelist))
```
Razzeee: can you try to use variables that are not single characters?

tests/test_check_addon.py
mzfr: I got it from https://docs.python.org/dev/library/unittest.html#unittest.TestCase.setUp

tests/test_check_addon.py
mzfr: like should I change them within comprehension or should I change this line into multiple lines ?

mzfr:
@Razzeee I have rebased it and made the changes required.

PR #85, opened 2018-06-04
use requirements.txt file for dependencies

mzfr:
We are planning to use pipenv for maintaining dependencies but until then we'll be using requirements.txt file.
.travis.yml is updated for installing all the requirements from requirements.txt file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
 
 # command to install dependencies
 install:
-  - pip install pytest Pillow pycodestyle
+  - pip install -r requirements.txt
 # command to run tests
 script:
   - py.test -v
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,3 @@
+pillow
+requests
+radon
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -0,0 +1,5 @@
+pytest
```
Razzeee: Please remove pytest and pycodestyle as we only need them in the travis.yml

mzfr: about pytest and pycodestyle: we'll have to keep the pycodestyle in the file else we'll have error about "pycodestyle not installed" or something like that ?

mzfr: Done

.travis.yml
```diff
@@ -5,7 +5,7 @@ matrix:
 
 # command to install dependencies
 install:
-  - pip install pytest Pillow pycodestyle
+  - pip install -r requirements.txt
```
Razzeee: these are missing pytest and pycodestyle now

Razzeee: Hrm, seems to be included in the python image then. The docs even reference pytest https://docs.travis-ci.com/user/languages/python/
So we should be good

PR #86, opened 2018-06-04
Check complex entry point for xml file

### LINKED ISSUE #82: Entry point validation fails for resource addons
Resource addons don't name `.py` files but `.xml` files and thus radon can't parse them.

We should ignore those cases, but maybe check if the file it references does exist.

mzfr:
Fix #82 
I have created a new function for counting the lines of `.xml` file.
Now I have added a check that if extension of a file is `.xml` then it will be passed to the new functions i.e `number_of_line_for_xml` for counting the line.
Ony files with `.py` extension will be passed to radon for parsing.

**Note** 
I was not sure what should be the max_entry_point for xml file so I used the default one. Please let me know if we want to use another variable for it. Also should it be fixed value or should we pass it from config ?

### DIFF
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -347,11 +347,12 @@ def _check_complex_addon_entrypoint(report: Report, addon_path, parsed_xml, max_
             filepath = os.path.join(addon_path, library)
 
             if not os.path.isdir(filepath):
+                name, ext = os.path.splitext(filepath)
 
                 if os.path.exists(filepath):
-                    number_of_lines(report, filepath, library,
-                                    max_entrypoint_line_count)
-
+                    if ext == '.py':
+                        number_of_lines(report, filepath, library,
+                                        max_entrypoint_line_count)
                 else:
                     report.add(
                         Record(PROBLEM, "%s Entry point does not exists" % library))


kodi_addon_checker/check_addon.py
```diff
@@ -345,18 +345,34 @@ def _check_complex_addon_entrypoint(report: Report, addon_path, parsed_xml, max_
 
         if library:
             filepath = os.path.join(addon_path, library)
+            name, ext = os.path.splitext(filepath)
```
Razzeee: move this down into the next if

kodi_addon_checker/check_addon.py
```diff
@@ -345,18 +345,34 @@ def _check_complex_addon_entrypoint(report: Report, addon_path, parsed_xml, max_
...
 
             if not os.path.isdir(filepath):
 
-                if os.path.exists(filepath):
+                if os.path.exists(filepath) and ext == '.py':
                     number_of_lines(report, filepath, library,
                                     max_entrypoint_line_count)
-
+                elif ext == '.xml':
```
Razzeee: you still need `os.path.exists(filepath)` if you use an elif here

kodi_addon_checker/check_addon.py
```diff
@@ -345,18 +345,34 @@ def _check_complex_addon_entrypoint(report: Report, addon_path, parsed_xml, max_
...
+                    number_of_lines_for_xml(report, filepath, library,
```
Razzeee: I don't think we care about the size of an xml entry point at all. So you can drop this branch of the logic

Razzeee:
Only py files get compiled, so that's why we care about the complexity there. XML just has to be loaded as they are, so we just need to not run the check for these.

kodi_addon_checker/check_addon.py
mzfr: Done

kodi_addon_checker/check_addon.py
mzfr: Done

kodi_addon_checker/check_addon.py
```diff
@@ -347,10 +347,14 @@ def _check_complex_addon_entrypoint(report: Report, addon_path, parsed_xml, max_
...
             if not os.path.isdir(filepath):
+                name, ext = os.path.splitext(filepath)
 
                 if os.path.exists(filepath):
-                    number_of_lines(report, filepath, library,
-                                    max_entrypoint_line_count)
+                    if ext == '.py':
+                        number_of_lines(report, filepath, library,
+                                        max_entrypoint_line_count)
+                    elif ext == '.xml':
```
Razzeee: Remove this branch

kodi_addon_checker/check_addon.py
```diff
@@ -347,10 +347,14 @@ def _check_complex_addon_entrypoint(report: Report, addon_path, parsed_xml, max_
             filepath = os.path.join(addon_path, library)
 
             if not os.path.isdir(filepath):
+                name, ext = os.path.splitext(filepath)
 
                 if os.path.exists(filepath):
```
Razzeee: Please add an else condition where you basically say, that we can't find the file with that path and print the path.

Razzeee: Whoops somehow missed that in the diff

PR #87, opened 2018-06-04
Move all_repo_addons function out of __main__.py

mzfr:
So we were having some issue with having an import in `__init__.py` file. So now we are shifting the `_all_repo_addons` out of the `__main__.py` file. 
There will be no change in the behaviour of the code base


kodi_addon_checker/check_addon.py
```diff
@@ -457,3 +458,14 @@ def _check_for_existing_addon(report: Report, addon_path, all_repo_addons):
                 return
 
     report.add(Record(INFORMATION, "This is a new addon"))
+
+
+def _all_repo_addons():
```
Razzeee: this is now a public method, so please don't prefix it with `_`

PR #88, opened 2018-06-06
Update ignored list of pseudo addons.

### LINKED ISSUE #83: Update ignored list of pseudo addons for dependency detection
We will never be able to find these addons in any repo.

So let's ignore all of these:
https://github.com/xbmc/xbmc/blob/2f9feb2/xbmc/addons/AddonInfo.cpp#L39-L77

mzfr:
Fix #83

I have added all the names of pseudo addons to the ignore list

### DIFF
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -413,8 +413,17 @@ def _get_users_dependencies(parsed_xml):
 def _check_dependencies(report: Report, repo_addons, parsed_xml):
     """Check for any new dependencies in addon.xml file"""
     deps = _get_users_dependencies(parsed_xml)
-    ignore = ['xbmc.json', 'xbmc.gui', 'xbmc.json',
-              'xbmc.metadata', 'xbmc.python']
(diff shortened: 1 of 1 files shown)


kodi_addon_checker/check_addon.py
```diff
@@ -413,8 +413,18 @@ def _get_users_dependencies(parsed_xml):
 def _check_dependencies(report: Report, repo_addons, parsed_xml):
     """Check for any new dependencies in addon.xml file"""
     deps = _get_users_dependencies(parsed_xml)
-    ignore = ['xbmc.json', 'xbmc.gui', 'xbmc.json',
-              'xbmc.metadata', 'xbmc.python']
+    ignore = ['xbmc.json', 'xbmc.gui', 'xbmc.metadata', 'xbmc.python', "xbmc.metadata.scraper.albums",
```
Razzeee: your mixing `"` and `'` here. Settle for one :)
I would go with `'` I think.

And maybe even drop the old ones, as they are not in the reference list

Razzeee: your mixing `"` and `'` here. Settle for one :)
I would go with `'` I think.

And maybe even drop the old ones, as they are not in the reference list

PR #92, opened 2018-06-10
catch exceptions per addon for a repo check

mzfr:
I have added multiple error as just using bare except was too vague and possibly we can expect multiple exceptions from our checks.

### DIFF
--- a/kodi_addon_checker/check_repo.py
+++ b/kodi_addon_checker/check_repo.py
@@ -1,6 +1,6 @@
 import os
 import kodi_addon_checker.check_addon as check_addon
-from kodi_addon_checker.record import INFORMATION, Record
+from kodi_addon_checker.record import INFORMATION, PROBLEM, Record
 from kodi_addon_checker.report import Report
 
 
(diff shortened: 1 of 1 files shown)


kodi_addon_checker/check_repo.py
```diff
@@ -12,7 +12,9 @@ def check_repo(repo_path, branch_name, all_repo_addons, pr, config):
...
         if addon_folder[0] != '.':
             addon_path = os.path.join(repo_path, addon_folder)
-            addon_report = check_addon.start(addon_path, branch_name, all_repo_addons, pr, config)
-            repo_report.add(addon_report)
-
+            try:
+                addon_report = check_addon.start(addon_path, branch_name, all_repo_addons, pr, config)
+                repo_report.add(addon_report)
+            except (IOError, SyntaxError, RuntimeError, TypeError):
+                repo_report.add(Record(PROBLEM, "There is some issue with your addon"))
```
Rechi: @Razzeee do we want to blame the addon?
can the backtrace be reported somewhere?

mzfr: If not the addon then maybe something general like `Something went wrong` or something along that line.

Rechi: yes, as long as it counted as error/problem
why not catch all exceptions?

mzfr: By all exception you mean bare except i.e `except exception as e` ?

PR #95, opened 2018-06-13
Remove unused variable and import

mzfr:
Remove Unused import logger from check_addon.py
Remove Unused variable 'importer' from common.py


Rechi:
I would have made it two commits (the and in the commit message indicates that).

Issue #96, opened 2018-06-23
Add script.module.pil to dependency ignore


Razzeee:

mzfr:
@Razzeee this issue is fixed

PR #97, opened 2018-06-24
Update dependency ignore list

mzfr:
Fix for issue #96

### DIFF
--- a/kodi_addon_checker/check_dependencies.py
+++ b/kodi_addon_checker/check_dependencies.py
@@ -2,27 +2,31 @@ from distutils.version import LooseVersion
 from .report import Report
 from .record import PROBLEM, Record, WARNING
 
+common_ignore_deps = ['xbmc.metadata.scraper.albums', 'xbmc.metadata.scraper.albums', 'xbmc.metadata.scraper.movies',
+                      'xbmc.metadata.scraper.musicvideos', 'xbmc.metadata.scraper.tvshows',
+                      'xbmc.metadata.scraper.library', 'xbmc.ui.screensaver', 'xbmc.player.musicviz',
+                      'xbmc.python.pluginsource', 'xbmc.python.script', 'xbmc.python.weather', 'xbmc.python.lyrics',
+                      'xbmc.python.library', 'xbmc.python.module', 'xbmc.subtitle.module', 'kodi.context.item',
+                      'kodi.game.controller', 'xbmc.gui.skin', 'xbmc.webinterface', 'xbmc.addon.repository',
+                      'xbmc.pvrclient', 'kodi.gameclient', 'kodi.peripheral', 'xbmc.addon.video',
+                      'xbmc.addon.audio', 'xbmc.addon.image', 'xbmc.addon.executable', 'kodi.addon.game',
+                      'kodi.audioencoder', 'kodi.audiodecoder', 'xbmc.service', 'kodi.resource.images',
+                      'kodi.resource.language', 'kodi.resource.uisounds', 'kodi.resource.games',
+                      'kodi.resource.font', 'kodi.inputstream', 'kodi.vfs', 'kodi.imagedecoder', 'xbmc.json',
+                      'xbmc.gui', 'xbmc.json', 'xbmc.metadata', 'xbmc.python', 'script.module.pil',
+                      'inputstream.adaptive', 'script.module.pycryptodome']
 
-def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml):
+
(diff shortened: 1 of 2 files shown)


enen92:
while you are at it maybe also add `inputstream.adaptive` and `script.module.pycryptodome`

Rechi:
inputstream.adaptive is only v17 and higher and script.module.pycryptodome is v18 only, not sure if that should be handled somehow.

Razzeee:
I would move the string we have right now to another method /file that accepts a branch and resolves it to a dependency list.

mzfr:
Oh so we can have a different dependency list for all the different branch ?

Razzeee:
Basically

mzfr:
@Razzeee  I have made the changes please take a look

kodi_addon_checker/check_dependencies.py
```diff
@@ -42,6 +47,21 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml):
...
 
+def _get_ignore_list(branch_name):
+
+    if branch_name == "leia":
+        common_ignore_deps.extend(["inputstream.adaptative", "inputstream.rtmp",
+                                   "script.module.cryptodome"])
+        return common_ignore_deps
+
+    elif branch_name == "krypton":
+        common_ignore_deps.extend(["inputstream.adaptative", "inputstream.rtmp"])
```
Razzeee: Not a big fan of mutating lists in this way. I would have used `+` I guess. See fore more info: https://stackoverflow.com/a/252711/2859813

mzfr: Just tested it and no it doesn't change with time.
Do you still want me to use the `+`

kodi_addon_checker/check_dependencies.py
```diff
@@ -2,26 +2,31 @@
...
+                      'xbmc.python.pluginsource', 'xbmc.python.script', 'xbmc.python.weather', 'xbmc.python.lyrics',
+                      'xbmc.python.library', 'xbmc.python.module', 'xbmc.subtitle.module', 'kodi.context.item',
+                      'kodi.game.controller', 'xbmc.gui.skin', 'xbmc.webinterface', 'xbmc.addon.repository',
+                      'xbmc.pvrclient', 'kodi.gameclient', 'kodi.peripheral', 'xbmc.addon.video',
+                      'xbmc.addon.audio', 'xbmc.addon.image', 'xbmc.addon.executable', 'kodi.addon.game',
+                      'kodi.audioencoder', 'kodi.audiodecoder', 'xbmc.service', 'kodi.resource.images',
+                      'kodi.resource.language', 'kodi.resource.uisounds', 'kodi.resource.games',
+                      'kodi.resource.font', 'kodi.inputstream', 'kodi.vfs', 'kodi.imagedecoder', 'xbmc.json',
+                      'xbmc.gui', 'xbmc.json', 'xbmc.metadata', 'xbmc.python', 'script.module.pil',
+                      'inputstream.adaptive', 'script.module.pycryptodome']
```
Rechi: shouldn't be added here

kodi_addon_checker/check_dependencies.py
```diff
@@ -42,6 +47,21 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml):
                                   % (required_addon, required_version, available_version)))
 
 
+def _get_ignore_list(branch_name):
+
+    if branch_name == "leia":
+        common_ignore_deps.extend(["inputstream.adaptative", "inputstream.rtmp",
```
Rechi: `inputstream.adaptive`

kodi_addon_checker/check_dependencies.py
Rechi: `inputstream.adaptive`

kodi_addon_checker/check_dependencies.py
Razzeee: https://github.com/xbmc/addon-check/commit/862de90c1ba3b7c1afff310af92802979cb32bb0

Rechi: `script.module.pycryptodome` is also wrong here

kodi_addon_checker/check_dependencies.py
```diff
@@ -42,6 +47,21 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml):
...
+                                   "script.module.cryptodome"])
```
Rechi: `script.module.pycryptodome`

PR #98, opened 2018-06-29
Add pylint checks

mzfr:
With this we will be able check our addons with pylint.
The output will look like:

| File name            | Line no  | Message                                   |
| ---------------------|----------|-------------------------------------------|      
|kodi.py               | 24       | Imports from package xbmc are not grouped |                                                                  
| url_dispatcher.py    | 99       | Line too long (165/120)                   |                                                                                    

Currently all the checks are enabled on line 17 in `.pylintrc` but we can then disable the ones we don't want.


kodi_addon_checker/pylint_checks.py
```diff
@@ -0,0 +1,31 @@
...
+
+def analyze(report: Report, filename: str):
+
+    ARGS = ["-r", "n", "--score=yes", "--lint-all=y", "--rcfile=kodi_addon_checker/.pylintrc"]
+    out = io.StringIO()
+    Run([filename] + ARGS, reporter=JSONReporter(out), do_exit=False)
+
+    json_data = json.loads(out.getvalue())
+    for issue in json_data:
+        table_format.append([short_path(issue['path']), issue['line'], issue['message']])
```
enen92: Can we extract the issue code number (e.g `E0601`) from the issue? It might be a good idea to prepare the checker for some post-processing on the collected issues before returning. Example, pylint will report `import routing` as a wrong import as the module can't be found in site-packages but the dev might be importing it as a script module in addon.xml. This will lead to a lot of false positives.

kodi_addon_checker/pylint_checks.py
```diff
@@ -0,0 +1,31 @@
...
+
+    table = AsciiTable(table_format).table
+    report.add(Record(WARNING, table))
```
enen92: I don't think this is the correct approach. The linter will report issues with different degrees of severity so the tool cannot simply return a warning with all the issues in a table. Ideally the tool should either calculate a "quality gate" or at least changing the record level accordingly. For example, in pylint, `Exxxx` messages mean `(syntax error raised for a module) and `Fxxxx` messages mean `(error prevented analysis)`. If pylint finds any of those messages, the reporter should log it as `PROBLEM` instead of warning

mzfr: we are still to decide what type of errors need to be caught. 
In this we are doing checks for all the pylint codes. But once we are final about the codes then I'll make the changes accordingly

kodi_addon_checker/check_addon.py
```diff
@@ -42,6 +43,10 @@ def start(addon_path, branch_name, all_repo_addons, pr, config=None):
 
             check_dependencies.check_addon_dependencies(addon_report, repo_addons, parsed_xml, branch_name)
 
+            pylint_checks.analyze(addon_report, addon_path)
+
+            check_dependencies.check_addon_dependencies(addon_report, repo_addons, parsed_xml)
```
Rechi: added by accident (already present above with the correct arguments)

requirements.txt
```diff
@@ -1,3 +1,5 @@
 pillow
 requests
 radon
+pylint
```
Rechi: `git+git://github.com/mzfr/pylint.git@master`

kodi_addon_checker/.pylintrc
```diff
@@ -0,0 +1,42 @@
+[Master]
+
+#Files to be ignored
+ignore=CVS,ini,xml
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# Specify a configuration file.
+rcfile=/home/mzfr/dev/Addon-check/Misc/Demo/
```
Rechi: remove

PR #99, opened 2018-07-06
Add support for pipenv

mzfr:
I have removed the `requirements.txt` file and now anyone can use pipfiles for installing dependencies. 

Currently there are only three packages i.e (`pillow`, `radon`, `requests`) present in pipfile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,7 @@ matrix:
 
 # command to install dependencies
 install:
-  - pip install -r requirements.txt
+  - make init
 
 before_script:
   - if [[ -n "$REPO" && -n "$BRANCH" ]]; then
@@ -91,7 +91,7 @@ script:
       python3 -m kodi_addon_checker --branch=$BRANCH ../repo;
     fi
   - if [[ -z "$REPO" || -z "$BRANCH" ]]; then
-      py.test -v;
+      make test;
     fi
   - if [[ -z "$REPO" || -z "$BRANCH" ]]; then
       pycodestyle . --max-line-length=120;
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,6 @@
+init:
+	pip install pipenv
+	pipenv install
+
+test: init
+	pipenv run pytest tests/.
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,15 @@
+[[source]]
+url = "https://pypi.org/simple"
+name = "pypi"
+verify_ssl = true
+
+[packages]
+requests = "*"
+radon = "*"
+Pillow = "*"
+
+[requires]
+python_version = "3.5"
+
+[dev-packages]
+pytest = "*"
--- a/setup.py
+++ b/setup.py
@@ -2,9 +2,6 @@ import os
 import setuptools
 
 
-with open('requirements.txt') as f:
-    requirements = f.read().splitlines()
-
 _ROOT = os.path.abspath(os.path.dirname(__file__))
 
 with open(os.path.join(_ROOT, 'README.md')) as f:
@@ -20,7 +17,7 @@ setuptools.setup(
     url="https://github.com/xbmc/addon-check",
     download_url="https://github.com/xbmc/addon-check/archive/master.zip",
     packages=setuptools.find_packages(exclude=['script.test', 'tests*']),
-    install_requires=requirements,
+    install_requires=['pillow', 'requests', 'radon'],
     setup_requires=['setuptools>=38.6.0'],
     entry_points={'console_scripts': [
         'kodi-addon-checker = kodi_addon_checker.__main__:main']},
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,0 +1,230 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "269fc39c0c2ec7190b52193bc2d5cfd83404a85d00a653672331eb1131f2a4c0"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.5"
+        },
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.org/simple",
+                "verify_ssl": true
+            }
+        ]
+    },
+    "default": {
+        "certifi": {
+            "hashes": [
+                "sha256:13e698f54293db9f89122b0581843a782ad0934a4fe0172d2a980ba77fc61bb7",
+                "sha256:9fa520c1bacfb634fa7af20a76bcbd3d5fb390481724c597da32c719a7dca4b0"
+            ],
+            "version": "==2018.4.16"
+        },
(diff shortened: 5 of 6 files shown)


Makefile
```diff
@@ -0,0 +1,6 @@
+init:
+	pip install pipenv
+	pipenv install
+
+test:
+	pipenv run pytest tests/.
```
Razzeee: I have no idea about make files, but I would have expected this to be a step in the travis config

Pipfile
```diff
@@ -0,0 +1,14 @@
...
+
+[packages]
+requests = "*"
+radon = "*"
+Pillow = "*"
+
+[requires]
+python_version = "3.5"
+
+[dev-packages]
```
Razzeee: Install `pytest` here

.travis.yml
```diff
@@ -77,8 +77,7 @@ matrix:
       env: REPO=repo-skins BRANCH=leia
 
 # command to install dependencies
-install:
-  - pip install -r requirements.txt
+install: "make"
```
Rechi: Has to be squashed into the commit where `requirements.txt` is removed or committed before that.
Any reason to change style?
```
install:
  - make
```

Makefile
```diff
@@ -0,0 +1,6 @@
+init:
+	pip install pipenv
+	pipenv install
+
+test:
```
Rechi: Doesn't test target need the pipenv from the init target?

Makefile
mzfr: Instead of running it from .travis.yml it will be handled by make file

.travis.yml
mzfr: Okay then I'll have to squash the last three commits.

No reason for style change. Will make it the old way.

Makefile
mzfr: I don't get it. What do mean by `init target` ?

Rechi: I assume you have to execute `pip install pipenv` & `pipenv install` that you can run `pipenv run pytest tests/.`, I'm correct?

mzfr: but I am doing that right ?
I mean the first two lines are doing the installation and then in `test` section I am running the tests

Rechi: If you run `make test`, the commands from `init` target won't get executed.

.travis.yml
```diff
@@ -78,7 +78,7 @@ matrix:
 
 # command to install dependencies
 install:
-  - pip install -r requirements.txt
+  - make
```
Rechi: `make init`, to make it explicit which target this make call currently executes

Makefile
mzfr: but I am not doing it I mean in .travis.yml I am doing just `install: -make` to run the whole Makefile

Rechi: - you don't execute both targets by calling `make`
- I expect that running test also sets up the environment before executing the tests

mzfr: oh yes just noticed that so I would have to change the `install: make` to

mzfr: oh yes just noticed it so now I'll have to change `install: make` to
```
install:
    - make init
    - make test
```

Rechi: two questions with this change now
- Why do you have to call init before test (if you want to do both in the same step)? Because your Makefile is wrong.
- Why run test in install step?

mzfr: * How is my makefile wrong ?

Rechi: test doesn't have init as dependency.

mzfr: okay now I got. so it should look like:

mzfr: okay so now I got it. It should look like:

`Makefile`
```
init:
    pip install pipenv
    pipenv install 

test:
    pipenv run pytest tests/.
```

`.travis.yml`
```
install:
    - make

sript:
    - make test
```

right ?

Rechi: - travis looks correct, beside that you should keep the additional repo tests
- test target still doesn't depend on init target (you haven't changed anything)

mzfr: * yes those additional repo test will be there
* by test target do you mean instead of `tests/.` I should use just `tests` like  `pipenv run pytest tests`

Rechi: I like having a makefile. @Razzeee it seems you also don't get it.
```
init:
	pip install pipenv
	pipenv install

test: init
	pipenv run pytest tests/.
```

mzfr: All of what I said was from https://docs.pipenv.org/advanced/#travis-ci

Rechi: It's correct for running it with travis, but you may want to run `make test` locally too and don't want to bother running `make init` explicitly first.

mzfr: But even without you're method when I do `make test` they seems to be running. fine.

Razzeee: Yes, for local it's fine. But having it like that will make travis do it twice, if we don't omit the init in that file.

Which also leads to me asking myself if there is a cache/performance gain to doing it in the install step.

Rechi: Then it has to be further optimized.

.travis.yml
```diff
@@ -87,6 +87,7 @@ before_script:
 
 # command to run tests
 script:
+  - make test
```
Rechi: I don't think this has to run in all 19 jobs

mzfr:
Then  for what all jobs should I do it ?

.travis.yml
mzfr: Then for what all jobs should I do it ?

Rechi: just for the default one

setup.py
```diff
@@ -20,7 +17,7 @@
     url="https://github.com/xbmc/addon-check",
     download_url="https://github.com/xbmc/addon-check/archive/master.zip",
     packages=setuptools.find_packages(exclude=['script.test', 'tests*']),
-    install_requires=requirements,
+    install_requires=['pillow', 'requests', 'radon'],
```
Rechi: Why hardcore the dependencies here instead of keeping the `requirements.txt` file?

mzfr: I just thought that if we are using pipenv then we can get rid of the requirements.txt file([can be problematic](https://www.kennethreitz.org/essays/a-better-pip-workflow)). Plus it actually doesn't effect anything in wrong way. Does it ?

Rechi: I don't like that dependencies have now to be changed in two places.
Moving one of the places somewhere into a large file makes it even worse.
Can't you read the required dependencies from Pipfile?

mzfr:
We shouldn't do that .
See : https://docs.pipenv.org/advanced/#pipfile-vs-setuppy

Razzeee:
If we refer to the author of pipenv, we need to question if it even makes sense for us to switch. https://github.com/pypa/pipenv/issues/1161#issuecomment-349972287

Originally I was thinking we qualify as an application, but due to us distributing via pip it seems like we qualify more in the library sense.

Razzeee:
I don't think we want to do this right now, so closing this.

PR #100, opened 2018-07-11
Add check for Python3 compatibility of addons

mzfr:
This will help in detecting the python2 code in addons submitted. 

Currently the output format for this is sloppy.

### DIFF
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -14,6 +14,7 @@ from . import check_entrypoint
 from . import handle_files
 from . import check_files
 from . import check_string
+from . import check_py3_compatibility
 
 REL_PATH = ""
 ROOT_URL = "http://mirrors.kodi.tv/addons/{branch}/addons.xml.gz"
@@ -54,6 +55,9 @@ def start(addon_path, branch_name, all_repo_addons, pr, config=None):
                 "max_entrypoint_count", 15)
             check_entrypoint.check_complex_addon_entrypoint(
                 addon_report, addon_path, parsed_xml, max_entrypoint_count)
+
+            if branch_name not in ['gotham', 'helix', 'isengard', 'jarvis']:
+                check_py3_compatibility.Check_Py3_compatibility(addon_report, addon_path)
 
             if config.is_enabled("check_license_file_exists"):
                 # check if license file is existing
--- a/kodi_addon_checker/check_py3_compatibility.py
+++ b/kodi_addon_checker/check_py3_compatibility.py
@@ -1,0 +1,60 @@
+import os
+
+from lib2to3 import refactor
+from tabulate import tabulate
+
+from .report import Report
+from .record import Record, INFORMATION
+
+
+class KodiRefactoringTool(refactor.RefactoringTool):
+
+    def __init__(self, report, *args, **kwargs):
+        self.report = report
+        super(KodiRefactoringTool, self).__init__(*args, **kwargs)
+
+    def short_path(self, path):
+        return os.path.split(path)[1]
+
+    def print_output(self, old, new, filepath, equal):
+        """
+        Called with the old version, new version, and filepath of a
+        refactored file.
+
+            :old: old text of refactored file
+            :new: new text of refactored file
+            :filepath: Path of the file
+            :equal: Tells whether or not old is equal to new
+        """
+
+        self.headers = ['#', 'Existing Code', 'Changes required']
+        self.table = []
+
+        if equal:
+            return
+
+        required_changes = new.splitlines()
+        existing_line = old.splitlines()
+
+        for line in range(min(len(required_changes), len(existing_line))):
+            if existing_line[line] != required_changes[line]:
+                self.table.append([line + 1, existing_line[line], required_changes[line]])
+
+        self.output = tabulate(self.table, headers=self.headers, tablefmt='fancy_grid')
+        self.report.add(Record(INFORMATION, self.short_path(filepath) + '\n' + self.output))
+
+
+def Check_Py3_compatibility(report: Report, path: str):
+    """
+     Checks compatibility of addons with python3
+        :path: path to the addon
+    """
+    fixer_names = []
+    list_of_fixes = ['import', 'dict', 'except', 'filter', 'has_key', 'itertools',
+                     'map', 'ne', 'next', 'print', 'renames', 'types', 'xrange', 'zip']
+
+    for fix in list_of_fixes:
+        fixer_names.append('lib2to3.fixes.fix_' + fix)
+
+    rt = KodiRefactoringTool(report, fixer_names, options=None, explicit=None)
+    rt.refactor([path])
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 pillow
 requests
 radon
+tabulate


kodi_addon_checker/check_addon.py
```diff
@@ -55,6 +56,9 @@ def start(addon_path, branch_name, all_repo_addons, pr, config=None):
             check_entrypoint.check_complex_addon_entrypoint(
                 addon_report, addon_path, parsed_xml, max_entrypoint_count)
 
+            if branch_name == "leia":
```
Rechi: no `branch_name` checks

mzfr: I thought @Razzeee said we should run this only for branches above `leia`

kodi_addon_checker/check_py3_compatibility.py
```diff
@@ -0,0 +1,83 @@
...
+        Called with the old version, new version, and filepath of a
+        refactored file.
+
+            :old: old text of refactored file
+            :new: new text of refactored file
+            :filepath: Path of the file
+            :equal: Tells whether or not old is equal to new
+        """
+
+        self.table_format = [['File', 'Line-No', 'Exisiting Code', 'Changes requried']]
```
Rechi: typos: `Existing` and `required`

kodi_addon_checker/check_py3_compatibility.py
```diff
@@ -0,0 +1,83 @@
...
+            self.report.add(Record(INFORMATION, self.short_path(filepath)))
+
+            if bool(set(new).intersection(old)):  # Tell whether there is some diff or not
+                new_changes = set(new.splitlines()) - set(old.splitlines())  # Give the requried changes
+                existing_line = set(old.splitlines()) - set(new.splitlines())  # Gives existing line
+                existing_code = self.line_num_for_phrase(filepath, list(existing_line))
+
+                for index, change in enumerate(sorted(existing_code)):
+                    self.table_format.append([self.short_path(filepath), existing_code[change], change,
+                                              sorted(new_changes)[index]])
```
Rechi: strip `Existing Code` and `Changes required` column to not bloat the table with useless white-spaces

kodi_addon_checker/check_addon.py
Rechi: That's not what your code does.
IMO it should be run on all branches (or at least kyrpton and higher), but type of report (info and error) should be different.

kodi_addon_checker/check_addon.py
```diff
@@ -55,6 +56,9 @@ def start(addon_path, branch_name, all_repo_addons, pr, config=None):
             check_entrypoint.check_complex_addon_entrypoint(
                 addon_report, addon_path, parsed_xml, max_entrypoint_count)
 
+            if branch_name >= "krypton":
```
Rechi: This is still a hack. Why not make it correct by adding a config setting?

kodi_addon_checker/check_py3_compatibility.py
```diff
@@ -0,0 +1,84 @@
...
+            :new: new text of refactored file
+            :filepath: Path of the file
+            :equal: Tells whether or not old is equal to new
+        """
+        tabulate.WIDE_CHARS_MODE = False
+        self.headers = ['File', 'Line-No', 'Existing Code', 'Changes required']
+        self.table = []
+
+        if not equal:
+            self.report.add(Record(INFORMATION, self.short_path(filepath)))
```
Rechi: either log file path septate or in the table, but once is enough

kodi_addon_checker/check_py3_compatibility.py
```diff
@@ -0,0 +1,84 @@
...
+                new_changes = set(new.splitlines()) - set(old.splitlines())  # Give the requried changes
+                existing_line = set(old.splitlines()) - set(new.splitlines())  # Gives existing line
+                existing_code = self.line_num_for_phrase(filepath, list(existing_line))
+
+                for index, change in enumerate(sorted(existing_code)):
+                    self.table.append([self.short_path(filepath), existing_code[change], change,
+                                       sorted(new_changes)[index]])
+
+        self.output = tabulate(self.table, headers=self.headers, tablefmt='fancy_grid')
+        self.report.add(Record(INFORMATION, self.output))
```
Rechi: this has to go into the `if not equal` too or add the following at the begin of the function
```
if equal:
    return
```

mzfr: Fixed

kodi_addon_checker/check_py3_compatibility.py
```diff
@@ -0,0 +1,85 @@
...
+            :filepath: Path of the file
+            :equal: Tells whether or not old is equal to new
+        """
+        tabulate.WIDE_CHARS_MODE = False
+        self.headers = ['File', 'Line-No', 'Existing Code', 'Changes required']
+        self.table = []
+        if equal:
+            return
+
+        if not equal:
```
Rechi: not needed anymore with the early return

kodi_addon_checker/check_addon.py
```diff
@@ -55,6 +56,9 @@ def start(addon_path, branch_name, all_repo_addons, pr, config=None):
             check_entrypoint.check_complex_addon_entrypoint(
                 addon_report, addon_path, parsed_xml, max_entrypoint_count)
 
+            if config.is_enabled("check_py3_compatibility"):
```
Razzeee: We should make all these settings depend on the version branch. And remove the config as far as we can. Or maybe have a proxy config that is inlined in the tool, which resolves branches to individual config settings

requirements.txt
```diff
@@ -1,3 +1,5 @@
 pillow
 requests
 radon
+wcwidth
```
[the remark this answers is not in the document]
mzfr: > Wide character support is enabled automatically if wcwidth library is already installed. To disable wide characters support without uninstalling wcwidth, set the global module-level flag WIDE_CHARS_MODE:
```
  import tabulate
  tabulate.WIDE_CHARS_MODE = False
```

This was done for handling the width of the columns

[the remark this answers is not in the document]
Rechi: I don't know if wide character support is needed, but if then add `tabulate[widechars]` only in this file.

See https://pip.pypa.io/en/latest/reference/pip_install/#examples (point 6) and https://bitbucket.org/astanin/python-tabulate/src/master/setup.py#setup.py-60

kodi_addon_checker/check_py3_compatibility.py
```diff
@@ -0,0 +1,82 @@
...
+            :equal: Tells whether or not old is equal to new
+        """
+        tabulate.WIDE_CHARS_MODE = False
+        self.headers = ['File', 'Line-No', 'Existing Code', 'Changes required']
+        self.table = []
+
+        if equal:
+            return
+
+        if bool(set(new).intersection(old)):  # Tell whether there is some diff or not
```
Rechi: Isn't that exactly the same what the equal parameter already tells?

kodi_addon_checker/check_py3_compatibility.py
```diff
@@ -0,0 +1,82 @@
...
+            new_changes = set(new.splitlines()) - set(old.splitlines())  # Give the requried changes
+            existing_line = set(old.splitlines()) - set(new.splitlines())  # Gives existing line
+            existing_code = self.line_num_for_phrase(filepath, list(existing_line))
+
+            for index, change in enumerate(sorted(existing_code)):
+                self.table.append([self.short_path(filepath), existing_code[change], change,
+                                   sorted(new_changes)[index]])
+
+        self.output = tabulate(self.table, headers=self.headers, tablefmt='fancy_grid')
+        self.report.add(Record(INFORMATION, self.output))
```
Rechi: `, '\n' + self.output))` so that first line of the table isn't indented by `INFO: `

kodi_addon_checker/check_py3_compatibility.py
mzfr: When I use pdb to check they have different values. The time `equal` was False but the `if bool(set(new).intersection(old))` returns true

requirements.txt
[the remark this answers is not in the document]
mzfr: I think I'll remove this because after testing on multiple addons I don't see any changes in the table

kodi_addon_checker/check_py3_compatibility.py
Rechi: Please give me an example addon where that happens.

mzfr: Ex: `plugin.video.twitch`

I did `import pdb; pdb.set_trace()` on line above `if equal:` and when the pdb console comes up 
I did `print(equal)` which gives `False` and then doing `bool(set(new).intersection(old))` gives true

kodi_addon_checker/check_py3_compatibility.py
```diff
@@ -0,0 +1,82 @@
...
+            new_changes = set(new.splitlines()) - set(old.splitlines())  # Give the requried changes
+            existing_line = set(old.splitlines()) - set(new.splitlines())  # Gives existing line
+            existing_code = self.line_num_for_phrase(filepath, list(existing_line))
+
+            for index, change in enumerate(sorted(existing_code)):
+                self.table.append([existing_code[change], change, sorted(new_changes)[index]])
+
+        self.report.add(Record(INFORMATION, self.short_path(filepath)))
+        self.output = tabulate(self.table, headers=self.headers, tablefmt='fancy_grid')
+        self.report.add(Record(INFORMATION, '\n' + self.output))
```
Rechi: `..., self.short_path(filepath) + '\n' + self.output))` and remove the separate report of the file path

IMO this should be at least a warning

kodi_addon_checker/check_py3_compatibility.py
```diff
@@ -0,0 +1,82 @@
...
+        Called with the old version, new version, and filepath of a
+        refactored file.
+
+            :old: old text of refactored file
+            :new: new text of refactored file
+            :filepath: Path of the file
+            :equal: Tells whether or not old is equal to new
+        """
+
+        self.headers = ['Line-No', 'Existing Code', 'Changes required']
```
Rechi: you could safe some column with space by renaming `Line-No` to `Line`

kodi_addon_checker/check_py3_compatibility.py
Rechi: I thought a bit more about `bool(set(new).intersection(old))` and it doesn't do what you expect.
Try to understand what it really and then you will notice you could also replace it with True.

Your observation doesn't show anything right or wrong. What would you expect?

kodi_addon_checker/check_py3_compatibility.py
```diff
@@ -0,0 +1,80 @@
...
+        for index, change in enumerate(sorted(existing_code)):
+            self.table.append([existing_code[change], change, sorted(new_changes)[index]])
+
+        self.output = tabulate(self.table, headers=self.headers, tablefmt='fancy_grid')
+        self.report.add(Record(INFORMATION, self.short_path(filepath) + '\n' + self.output))
+
+
+def Check_Py3_compatibility(report: Report, path: str):
+    """
+     Checks compatinilty of addons with python3
```
Razzeee: Typo

mzfr: Fixed

Razzeee:
Looking good so far, but we might want to do linebreaks in the table after some characters. As it get's way to wide to read.

image


Also this has interesting markup, this one might just be due to travis
image

PR #101, opened 2018-07-15
print reports instantly

Rechi:


kodi_addon_checker/plugins/testarray_reporter.py
```diff
@@ -0,0 +1,12 @@
+from kodi_addon_checker.report import Record
+from kodi_addon_checker.reporter import Reporter, reporter
+
+
+@reporter(name="testarray", enabled=False)
+class TestArrayReporter(Reporter):
```
Razzeee: Not 100% happy with Test being in this name.

Razzeee: I would be fine with just dropping test from the name.

PR #103, opened 2018-08-05
Add check for xbmc.python version

mzfr:
This will check for the version of `xbmc.python` in addon.xml file according to the kodi version

### DIFF
--- a/kodi_addon_checker/check_dependencies.py
+++ b/kodi_addon_checker/check_dependencies.py
@@ -1,4 +1,6 @@
+import logging
 from distutils.version import LooseVersion
+
 from .report import Report
 from .record import PROBLEM, Record, WARNING
 
@@ -14,6 +16,11 @@ common_ignore_deps = ['xbmc.metadata.scraper.albums', 'xbmc.metadata.scraper.mov
                       'kodi.resource.language', 'kodi.resource.uisounds', 'kodi.resource.games',
                       'kodi.resource.font', 'kodi.inputstream', 'kodi.vfs', 'kodi.imagedecoder', 'xbmc.addon',
                       'xbmc.gui', 'xbmc.json', 'xbmc.metadata', 'xbmc.python', 'script.module.pil']
+
+VERSION_ATTRB = {'xbmc.python': {'gotham': '2.14.0', 'helix': '2.19.0', 'isengard': '2.20.0',
+                                 'jarvis': '2.24.0', 'krypton': '2.25.0', 'leia': '2.25.0'}}
+
+LOGGER = logging.getLogger(__name__)
 
 
 def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, branch_name: str):
@@ -32,6 +39,15 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
             if required_addon not in ignore:
                 report.add(Record(
                     PROBLEM, "Required addon %s not available in current repository." % required_addon))
+            elif required_addon in VERSION_ATTRB:
+                try:
+                    version = VERSION_ATTRB[required_addon][branch_name]
+                    if LooseVersion(version) != LooseVersion(required_version):
+                        report.add(Record(WARNING, "For %s it is advised to set %s version to %s" %
+                                          (branch_name, required_addon, required_version)))
+                except KeyError:
+                    LOGGER.warn("Misconfiguration in VERSION_ATTRB of check_dependencies")
+
         else:
             available_version = repo_addons[required_addon]
 
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ It can also be used locally for detecting problems in your addons.
 
 - Check if addon is compatible with python3 or not
 
+- Check specific [version attribute](https://kodi.wiki/view/Addon.xml#version_attribute_2)
+
 All of the validation and checks are done according to the kodi [addon rules](https://kodi.wiki/view/Add-on_rules)
 
 ## Installation


kodi_addon_checker/check_dependencies.py
```diff
@@ -26,12 +29,17 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
...
     deps = _get_users_dependencies(parsed_xml)
     ignore = _get_ignore_list(branch_name)
-
+    print(deps)
     for required_addon, required_version in deps.items():
         if required_addon not in repo_addons:
             if required_addon not in ignore:
                 report.add(Record(
                     PROBLEM, "Required addon %s not available in current repository." % required_addon))
+            elif required_addon == "xbmc.python":
```
Rechi: As already mentioned on slack, no specific version id hacks. Make this check general, no hard coded addon id  in the `check_addon_dependencies` function.

mzfr: General as in for all dependencies  i.e `xbmc.python`, `xbmc.gui` etc ?

Rechi: One has to be able to add another addon id and version in a list, instead of adding here another elif for further addons.

kodi_addon_checker/check_dependencies.py
```diff
@@ -32,6 +35,11 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
             if required_addon not in ignore:
                 report.add(Record(
                     PROBLEM, "Required addon %s not available in current repository." % required_addon))
+            else:
+                version = VERSION_ATTRB[required_addon][branch_name]
+                if not LooseVersion(version) == LooseVersion(required_version):
+                    report.add(Record(WARNING, "You are using and older version xbmc.python"))
```
enen92: 2 things:
- Don't hardcode `xbmc.python` here since the idea is to make it independent of the addon
- Can you report what version the author should use (i.e. "For {branch} it is advised to set {addon} version to {required_version}.

kodi_addon_checker/check_dependencies.py
```diff
@@ -32,6 +35,12 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
             if required_addon not in ignore:
                 report.add(Record(
                     PROBLEM, "Required addon %s not available in current repository." % required_addon))
+            else:
+                version = VERSION_ATTRB[required_addon][branch_name]
```
Razzeee: this can reference unknown keys, have you tested what happens then? Do we just crash?

kodi_addon_checker/check_dependencies.py
```diff
@@ -32,6 +35,12 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
...
+                if not LooseVersion(version) == LooseVersion(required_version):
+                    report.add(Record(WARNING, "For %s it is advised to set %s  version to %s" %
```
Razzeee: There is a space to much in the string

Razzeee: How about `For branch '%s' it is advised to set '%s' version to '%s'`

kodi_addon_checker/check_dependencies.py
mzfr: Fixed

kodi_addon_checker/check_dependencies.py
```diff
@@ -29,9 +32,15 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
 
     for required_addon, required_version in deps.items():
         if required_addon not in repo_addons:
-            if required_addon not in ignore:
+            if required_addon not in ignore and required_addon in VERSION_ATTRB:
```
Razzeee: `branch_name` might also be invalid. And this is not really a nice fix, as this would just swallow the error and nobody will ever know what it happend. :/

Razzeee: also this is the wrong way around I think. Best would be to extract this part of the logic into it's own method and write unit tests for this.

mzfr: How can `branch_name` might be invalid ?
I don't understand how this will swallow "error" when we won't have any error ? Can you please explain this with an example ?

> also this is the wrong way around I think. Best would be to extract this part of the logic into it's own method and write unit tests for this.

Meaning make a whole new function for this check ?

Razzeee: 1. It can't. We should be fine, unless we change https://github.com/xbmc/addon-check/blob/98b2954dddba6c7972e2a890c9e6f11b62235cc2/kodi_addon_checker/__main__.py#L56 and maybe add the next version. But forget about `VERSION_ATTRB` in the update process. Then we will be referencing something that doesn't exist and it will just crash, which is probably fine, but reporting errors is nicer most of the times.

2. Well if `required_addon in VERSION_ATTRB` is false we end up in the else branch. Which will just do it's thing and never mention that the required_addon was misconfigured by us somehow.

3. For this subpart, it's just a recommendation, as I've seen that you got it the wrong way around and writing tests helps figuring stuff like that out. But I would also understand if you think it's to difficult.

Branch name is just a string variable

mzfr: For 1 should report an error("There is some misconfiguration with your keys/values") in our format or should I log it ?

Razzeee: Log it please, it's an error on our (addon-check) site if it happens.

Rechi:
new feature has to be added in the readme

Rechi:
`Update README` isn't a descriptive commit message.

Rechi:
First line of the commit message (which you see by git log --oneline or on github in the commit log) still hasn't changed.

As soon as you give it a usefull name it will be similar to ’Add check for xbmc.python version’ and you should ask yourself, why is this a seperate commit.

kodi_addon_checker/check_dependencies.py
```diff
@@ -32,6 +39,15 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
             if required_addon not in ignore:
                 report.add(Record(
                     PROBLEM, "Required addon %s not available in current repository." % required_addon))
+            elif required_addon in VERSION_ATTRB:
+                try:
+                    version = VERSION_ATTRB[required_addon][branch_name]
+                    if LooseVersion(version) != LooseVersion(required_version):
+                        report.add(Record(WARNING, "For %s it is advised to set %s version to %s" %
+                                          (branch_name, required_addon, required_version)))
```
Rechi: This reports the wrong version which should be used.

mzfr: what do you mean by wrong version? like the version of the attribute ?

Rechi: e.g. `WARN: For helix it is advised to set xbmc.python version to 2.1.0`

PR #108, opened 2018-08-20
check transparency in screenshot

### LINKED ISSUE #30: Check screenshots for transparency

mzfr:
Fixes #30 
This is a simple check but if we want to check width/height of the screenshot or as said by @enen92 check aspect ratio then I'll make a new function for checking screenshot just like `_check_fanart` or `_check_icon`

### DIFF
--- a/kodi_addon_checker/check_artwork.py
+++ b/kodi_addon_checker/check_artwork.py
@@ -39,7 +39,6 @@ def _check_image_type(report: Report, image_type: str, parsed_xml, addon_path: s
             filepath = os.path.join(addon_path, image.text)
 
             if os.path.isfile(filepath):
-
                 report.add(Record(INFORMATION, "Image %s exists" % image_type))
                 try:
                     im = Image.open(filepath)
@@ -52,6 +51,8 @@ def _check_image_type(report: Report, image_type: str, parsed_xml, addon_path: s
                         _check_fanart(report, width, height)
                     else:
                         # screenshots have no size definitions
+                        if has_transparency(im):
+                            report.add(Record(PROBLEM, "%s should be solid. It has transparency." % image.text))
                         LOGGER.info("Artwork was a screenshot")
                 except IOError:
                     report.add(


kodi_addon_checker/check_artwork.py
```diff
@@ -51,7 +51,8 @@ def _check_image_type(report: Report, image_type: str, parsed_xml, addon_path: s
                     elif image_type == "fanart":
                         _check_fanart(report, width, height)
                     else:
-                        # screenshots have no size definitions
+                        if has_transparency(im):
+                            report.add(Record(PROBLEM, "screenshot.png should be solid. It has transparency."))
```
enen92: An addon can define multiple screenshots and only one have a transparency problem. Please report the correct file name instead of a generic `screenshot.png`

kodi_addon_checker/check_artwork.py
```diff
@@ -51,7 +51,8 @@ def _check_image_type(report: Report, image_type: str, parsed_xml, addon_path: s
                     elif image_type == "fanart":
                         _check_fanart(report, width, height)
                     else:
-                        # screenshots have no size definitions
```
enen92: no need to remove the comment, it is still valid since you don't introduce a check for that (and we don't know yet what to do with it). At least with the comment we know where to look and introduce the check.

kodi_addon_checker/check_artwork.py
Rechi: Icon and fanart is shown while browsing through addons in kodi. Screenshots are shown in the info screen of each addon.

Razzeee: image

kodi_addon_checker/check_artwork.py
```diff
@@ -52,6 +51,8 @@ def _check_image_type(report: Report, image_type: str, parsed_xml, addon_path: s
                         _check_fanart(report, width, height)
                     else:
                         # screenshots have no size definitions
+                        if has_transparency(im):
+                            report.add(Record(PROBLEM, "%s should be solid. It has transparency.", image.text))
```
Rechi: `... transparency." % image.text))`

PR #109, opened 2018-08-22
report all legacy strings.xml files

Rechi:


kodi_addon_checker/check_string.py
```diff
@@ -8,9 +10,9 @@ def check_for_legacy_strings_xml(report: Report, addon_path: str):
     """Find for the string.xml file in addon which was used in old versions
         :addon_path: path of the addon
     """
-    if handle_files.find_file_recursive("strings.xml", addon_path) is not None:
+    for file in handle_files.find_files_recursive("strings.xml", os.path.join(addon_path, "resources", "language")):
         report.add(
-            Record(PROBLEM, "Found strings.xml in folder %s please migrate to strings.po." % relative_path(addon_path)))
+            Record(PROBLEM, "Found %s please migrate to strings.po." % relative_path(file)))
```
enen92: Wondering if it makes sense to at least check the integrity of the xml files. Maybe not since you are already reporting it as a problem.

Rechi: https://github.com/xbmc/addon-check/blob/46b4168878499175e7f6da485ca745283e24991a/kodi_addon_checker/check_files.py#L11-L24

Razzeee:
Thank you! 🎉

PR #113, opened 2018-08-24
Add check for optional dependencies

### LINKED ISSUE #110: [RFC] Should an optional dependency be reported as a problem?
Ref: https://github.com/xbmc/repo-plugins/pull/2001

script.module.inputstreamhelper is an optional dependency of the youtube addon and only makes sense for v17+. Youtube is in the isengard repo and is able to operate even without inputstream.

Question is...should we report such cases as problem? IMO warning fits better if the dependency is optional.

mzfr:
Fixes #110.

used the existing function, `_get_users_dependencies`, to separate optional and required dependencies.

### DIFF
--- a/kodi_addon_checker/check_dependencies.py
+++ b/kodi_addon_checker/check_dependencies.py
@@ -2,7 +2,7 @@ import logging
 from distutils.version import LooseVersion
 
 from .report import Report
-from .record import PROBLEM, Record, WARNING
+from .record import PROBLEM, Record, WARNING, INFORMATION
 
 common_ignore_deps = ['xbmc.metadata.scraper.albums', 'xbmc.metadata.scraper.movies',
                       'xbmc.metadata.scraper.musicvideos', 'xbmc.metadata.scraper.tvshows',
@@ -31,8 +31,22 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
         :branch_name: name of the kodi branch/version
     """
 
-    deps = _get_users_dependencies(parsed_xml)
+    optional_deps, deps = _get_users_dependencies(parsed_xml)
     ignore = _get_ignore_list(branch_name)
+
+    if optional_deps:
+        for dependency, version in optional_deps.items():
+            if dependency not in repo_addons:
+                report.add(Record(INFORMATION,
+                                  "Optional dependency %s with version %s is not present in Kodi repository" %
+                                  (dependency, version)))
+            else:
+                deps_version = repo_addons[dependency]
+                if deps_version is not None:
+                    if LooseVersion(version) > LooseVersion(deps_version) and (dependency not in ignore):
+                        report.add(INFORMATION,
+                                   "version mismatch for optional dependencies: %s. Available: %s, Required: %s" %
+                                   (dependency, deps_version, version))
 
     for required_addon, required_version in deps.items():
         if required_addon not in repo_addons:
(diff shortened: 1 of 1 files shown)


kodi_addon_checker/check_dependencies.py
```diff
@@ -31,9 +31,14 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
         :branch_name: name of the kodi branch/version
     """
 
-    deps = _get_users_dependencies(parsed_xml)
+    optional_deps, deps = _get_users_dependencies(parsed_xml)
     ignore = _get_ignore_list(branch_name)
 
+    if optional_deps:
+        for dependency, version in optional_deps.items():
+            report.add(Record(INFORMATION, "%s with version %s is set as an optional dependency" %
```
enen92: Again, this does not add any value. The issue (INFO) in this case should only be reported if the dependency is not available in the repository. See the use case of the issue you're trying to solve. 

Setting dependencies as optional is per-si is an option for any addon...

kodi_addon_checker/check_dependencies.py
```diff
@@ -81,8 +95,12 @@ def _get_users_dependencies(parsed_xml):
     """Gets all the dependencies from a given addon
         :parsed_xml: parsed addon.xml
     """
+    user_deps = {}
+    optional_deps = {}
+    for i in parsed_xml.findall("requires/import"):
+        if i.get("otpional"):
```
anxdpanic: Typo. After correcting, I get the info report instead of the error when running locally.

mzfr: It was discussed that since this is a check for optional dependencies and an author would be aware of them(they require extra changes in addon.xml) so it would be nice to not make them error/warning but to keep it as info, just a nice suggestion for some improvement in addon.

anxdpanic: Makes sense, with the PR as is I still get the error. Changing i.get("otpional") to i.get("optional") I get the expected info report.

kodi_addon_checker/check_dependencies.py
```diff
@@ -31,9 +31,23 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
...
 
-    deps = _get_users_dependencies(parsed_xml)
+    optional_deps, deps = _get_users_dependencies(parsed_xml)
     ignore = _get_ignore_list(branch_name)
 
+    if optional_deps:
+        for dependency, version in optional_deps.items():
+            if dependency not in repo_addons:
+                report.add(Record(INFORMATION,
+                                  "Optional dependency %s with version %s is not present in XBMC repository" %
```
Razzeee: Kodi instead of XBMC

kodi_addon_checker/check_dependencies.py
```diff
@@ -31,9 +31,23 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
...
+                                  (dependency, version)))
+            else:
+                deps_version = repo_addons[dependency]
+                if deps_version is not None:
+                    if LooseVersion(version) > LooseVersion(deps_version) and (dependency not in ignore):
+                        report.add(INFORMATION,
+                                   "%s with have higher version requirements. Available: %s, Required: %s" %
```
Razzeee: `Version mismatch for optional dependency %s. Required: %s, Available: %s`

mzfr:
@Razzeee anything needed to be done in this?

kodi_addon_checker/check_dependencies.py
```diff
@@ -31,9 +31,23 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
...
+            if dependency not in repo_addons:
+                report.add(Record(INFORMATION,
+                                  "Optional dependency %s with version %s is not present in Kodi repository" %
+                                  (dependency, version)))
+            else:
+                deps_version = repo_addons[dependency]
+                if deps_version is not None:
+                    if LooseVersion(version) > LooseVersion(deps_version) and (dependency not in ignore):
+                        report.add(INFORMATION,
+                                   "version mismatch for optional dependencies. Available: %s, Required: %s" %
```
Razzeee: Would be nice to also print the dependency your testing for here.

PR #115, opened 2018-09-27
convert check_legacy_language_path config to branch checks

Rechi:
`.tests-config.json` isn't loaded from the current working directory, but from the addon or repo path. That causes that our PR builds aren't checked for legacy language paths.

### DIFF
--- a/tests/fixtures/.tests-config.json
+++ b/tests/fixtures/.tests-config.json
@@ -1,4 +1,4 @@
 {
-    "check_legacy_language_path": true,
+    "true_config": true,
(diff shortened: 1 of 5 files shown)


tests/fixtures/.tests-config.json
```diff
@@ -1,4 +1,4 @@
 {
-    "check_legacy_language_path": true,
+    "true_config": true,
```
enen92: what exactly is `true_config`, the name is not really clear to me. 
`config.is_enabled("true_config") is True`

PR #118, opened 2018-10-05
[schema] don't allow empty optional elements & limit news length

Rechi:


kodi_addon_checker/xml_schema/gotham_metadata.xsd
```diff
@@ -7,31 +7,37 @@
...
   </xs:element>
   <xs:simpleType name="simpleIdentifier">
     <xs:restriction base="xs:string">
       <xs:pattern value="(kodi|xbmc)\.addon\.metadata"/>
     </xs:restriction>
   </xs:simpleType>
+  <xs:simpleType name="nonEmptyString">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="\s*([^\s]\s*)+"/>
+      <xs:maxLength value="1500"/>
```
Razzeee: Is there a reason for 1500 or is this just what you figured would be a good size?

Rechi: Random size, tried a few numbers and looked what it'll break

PR #121, opened 2018-10-11
Change deps location to setup.py.

thiagoalmeidasa:
Instead of include requirements.txt in pypi package, we will handle
dependency management on setup.py.


Razzeee:
Thank you, but I will close this in favor of https://github.com/xbmc/addon-check/pull/122

PR #129, opened 2018-10-15
Add check for incorrectly set file permissions

### LINKED ISSUE #43: detect incorrectly set file permissions
Some accidentally set wrong file permissions while submitting

mzfr:
Fixes #43 
I have set this to be reported as a warning, let me know if we have to report it as `error` instead of `warning`.

### DIFF
--- a/kodi_addon_checker/check_files.py
+++ b/kodi_addon_checker/check_files.py
@@ -2,6 +2,7 @@ import os
 import re
 import json
 import xml.etree.ElementTree as ET
+from pathlib import Path
 from . import handle_files
 from .common import relative_path
 from .record import PROBLEM, Record, WARNING, INFORMATION
@@ -110,3 +111,16 @@ def check_file_whitelist(report: Report, file_index: list, addon_path: str):
                 report.add(Record(WARNING,
                                   "Found non whitelisted file ending in filename %s" %
                                   relative_path(os.path.join(file["path"], file["name"]))))
+
+
+def check_file_permission(report: Report, addon_path: str):
+    """Check whether the files present in addon are marked executable
+       or not
+        :addon_path: Path of the addon
+    """
+
+    files = Path(addon_path).glob('**/*')
+
+    for file in files:
+        if os.path.isfile(file) and os.access(str(file), os.X_OK):
+            report.add(Record(PROBLEM, "%s is marked as stand-alone executable" % relative_path(str(file))))
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ It can also be used locally for detecting problems in your addons.
 
 - Validate addon.xml against schemas
 
+- Check if files in addon are marked as executable or not.
+
 All of the validation and checks are done according to the kodi [addon rules](https://kodi.wiki/view/Add-on_rules)
 
 ## Installation
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -47,6 +47,8 @@ def start(addon_path, branch_name, all_repo_addons, pr, config=None):
 
             check_dependencies.check_addon_dependencies(addon_report, repo_addons, parsed_xml, branch_name)
 
+            check_files.check_file_permission(addon_report, addon_path)
+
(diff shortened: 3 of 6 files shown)


kodi_addon_checker/check_files.py
```diff
@@ -110,3 +111,16 @@ def check_file_whitelist(report: Report, file_index: list, addon_path: str):
...
                                   relative_path(os.path.join(file["path"], file["name"]))))
+
+
+def check_file_permission(report: Report, addon_path: str):
+    """Check whether the files present in addon are marked executable
+       or not
+        :addon_path: Path of the addon
+    """
+
+    files = list(Path(addon_path).glob('**/*.py'))
```
Rechi: no need to make the generator an explicit list

kodi_addon_checker/check_files.py
```diff
@@ -110,3 +111,16 @@ def check_file_whitelist(report: Report, file_index: list, addon_path: str):
...
+
+    for file in files:
+        if os.access(file, os.X_OK):
+            report.add(Record(WARNING, "%s is marked as stand-alone executable" % file))
```
Rechi: Why only `WARNING`? Is there any use case you have in mind?

Razzeee:
Can you write a test for this? Or are the params to hard to create/mock?

mzfr:
Test like passing an executable file and expecting an `error` right?

Razzeee:
Yes, but please also test the `happy path` when everything should be okay

kodi_addon_checker/check_files.py
```diff
@@ -110,3 +111,16 @@ def check_file_whitelist(report: Report, file_index: list, addon_path: str):
...
                                   relative_path(os.path.join(file["path"], file["name"]))))
+
+
+def check_file_permission(report: Report, addon_path: str):
+    """Check whether the files present in addon are marked executable
+       or not
+        :addon_path: Path of the addon
+    """
+
+    files = Path(addon_path).glob('**/*.py')
```
Rechi: Shouldn't the permission be checked for all files?
Anyone can imagine an use case where a non python file has to be marked executable?

Razzeee: No, your absolutely right.

kodi_addon_checker/check_files.py
```diff
@@ -110,3 +111,16 @@ def check_file_whitelist(report: Report, file_index: list, addon_path: str):
...
+def check_file_permission(report: Report, addon_path: str):
+    """Check whether the files present in addon are marked executable
+       or not
+        :addon_path: Path of the addon
+    """
+
+    files = Path(addon_path).glob('**/*')
+
+    for file in files:
+        if os.access(str(file), os.X_OK):
```
Rechi: Folders have to be marked executable as you otherwise don't have permissions to view it's content.

kodi_addon_checker/check_files.py
```diff
@@ -110,3 +111,16 @@ def check_file_whitelist(report: Report, file_index: list, addon_path: str):
...
+    """Check whether the files present in addon are marked executable
+       or not
+        :addon_path: Path of the addon
+    """
+
+    files = Path(addon_path).glob('**/*')
+
+    for file in files:
+        if os.path.isfile(file) and os.access(str(file), os.X_OK):
+            report.add(Record(PROBLEM, "%s is marked as stand-alone executable" % file))
```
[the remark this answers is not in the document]
Rechi: Would it be better to only report one error per addon with a list of files, instead of one error per file?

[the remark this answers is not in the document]
mzfr: @Razzeee should this be done?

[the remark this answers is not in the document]
Razzeee: Personally I like it this way around, reporting each as a single error.

But I agree on the relative path.

mzfr:
@razze I have made the suggested changes

Razzeee:
No, we're still not seeing relative paths with this PR.

mzfr:
sorry forgot to tweak the test with the changes. But now it is done, we can see the relative path with this PR.

Razzeee:
Thank you @mzfr

PR #130, opened 2018-10-15
check if settings.xml file exists or not

### LINKED ISSUE #127: Suggest to use new settings.xml layout for v18 and up
Since several version we support a new settings.xml format which is the same as Kodi itself uses.
As most bugs should have been solved by now we should suggest developers to start using this new layout.
I'm sure there is already some wiki page or forum post explaining how the new layout works

mzfr:
Fixes #127

### DIFF
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -41,6 +41,8 @@ def start(addon_path, branch_name, all_repo_addons, pr, config=None):
         if len(addon_xml.findall("*//broken")) == 0:
             file_index = handle_files.create_file_index(addon_path)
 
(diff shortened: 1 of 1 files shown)


Razzeee:
This is not what we want. `settings.xml` is optional, not mandatory.

What martijn is aksing for in that ticket is, that if this file exists, we check for the format of the xml. As there is an old format and a recommended new one.

MartijnKaijser:
Correct:
old format
https://github.com/xbmc/xbmc/blob/master/addons/metadata.themoviedb.org/resources/settings.xml
new format
https://github.com/xbmc/xbmc/blob/master/system/settings/settings.xml

PR #131, opened 2018-10-15
Ignore some hidden files and folders

### LINKED ISSUE #123: Don't process hidden folders
We should ignore hidden folders, everything starting with `.` like `.git`

See this log for some false alarms https://travis-ci.org/trakt/script.trakt/jobs/440324400

mzfr:
Fixes #123

### DIFF
--- a/kodi_addon_checker/handle_files.py
+++ b/kodi_addon_checker/handle_files.py
@@ -46,10 +46,14 @@ def create_file_index(path: str):
     """
     file_index = []
     for root, folders, files in os.walk(path, topdown=True):
+        if ".directory" in files:
+            files.remove(".directory")
         if ".git" in folders:
             folders.remove(".git")
         for file_name in files:
-            file_index.append({"path": root, "name": file_name})
+            if not os.path.basename(root).startswith("."):
+                file_index.append({"path": root, "name": file_name})
+
     return file_index
 
 


Rechi:
Better would be to not exclude all `.*` files/folders. Just exclude `.git` explecitly.

mzfr:
But why do we want to process any folder that is hidden?

Razzeee:
I think we're mixing stuff up.

If you run it on a single addon, local or in a git source repo, we don't
want to process hidden folders. Files should always be processed, only
exception is them being in a hidden folder.

If we're testing in the repo context those need to be reported and then
removed by the person sending the pr.

Rechi:
Repo vs single addon seperation is fine for me.

mzfr:
@Razzeee is there any changes required in this?

Razzeee:
@Rechi the PR flag get's handed down in both cases and is taken care of in a single check it executes at the moment.

I'm torn now, while I think me design above is correct, it would lead to you running the reporter local/in your git branch and not getting errors. Then sending it to our git and probably getting errors :/

Rechi:
@Razzeee what do you think about ignoring files which are marked as export-ignored in git (https://git-scm.com/docs/gitattributes#_creating_an_archive)?

Rechi:
It wouldn't ignore all files/folders starting with . by default, but you could exactly define which files/folders to ignore (eg also tests).

Razzeee:
I hoped to keep it convention over configuration :/

Rechi:
As long as the check run at xbmc/repo-* doesn't ignore any files, I would be okay with ignoring .* files.
But as you already have mentioned that will cause different results.

Razzeee:
I think we need two modes. Additional to PR. 

* Ignores those files, only used local and in developers repo
* Ignores nothing for our xbmc/repo-*

While it's not fun to get new errors when you open the PR version, it's better then having no error reporting at all. If we don't implement it, nobody will use it for their local/dev repos, as they will get false positives all the time.

kodi_addon_checker/handle_files.py
```diff
@@ -39,7 +39,8 @@ def create_file_index(path: str):
     file_index = []
     for dirs in os.walk(path):
         for file_name in dirs[2]:
-            file_index.append({"path": dirs[0], "name": file_name})
+            if not file_name.startswith("."):
```
Rechi: this only checks for hidden files, not folders

kodi_addon_checker/handle_files.py
```diff
@@ -37,9 +37,10 @@ def create_file_index(path: str):
         :path: path for the directory
     """
     file_index = []
-    for dirs in os.walk(path):
-        for file_name in dirs[2]:
-            file_index.append({"path": dirs[0], "name": file_name})
+    for root, _, files in os.walk(path):
+        for file_name in files:
+            if not file_name.startswith(".") and not os.path.basename(root).startswith("."):
```
Rechi: only ignores the files in hidden folders, not in non hidden sub-folders

kodi_addon_checker/handle_files.py
```diff
@@ -49,7 +49,9 @@ def create_file_index(path: str):
         if ".git" in folders:
             folders.remove(".git")
         for file_name in files:
-            file_index.append({"path": root, "name": file_name})
+            if not os.path.basename(root).startswith("."):
```
Rechi: only ignores the files in hidden folders, not in non hidden sub-folders

- .test/tesfile: ignored
- .test/test/tesfile: **not** ignored

Rechi: - `.test/tesfile`
  - root: `[...]/.test`
  - file_name = `testfile`
  - basename(root): `.test`
  - ".test".startswith("."): `True`
  - -> ignored
- `.test/test/tesfile`
  - root: `[...]/.test/test`
  - file_name = `testfile`
  - basename(root): `test`
  - "test".startswith("."): `False`
  - -> **not** ignored
- `test/.tesfile`
  - root: `[...]/test`
  - file_name = `.testfile`
  - basename(root): `test`
  - "test".startswith("."): `False`
  - -> **not** ignored

mzfr:
Now, this is what this code is doing:

Say we have the following directory structure:
```
├── .directory
├── folder1
│   ├── file.py
│   └── .sub_hidden1
└── .hidden
    ├── folder2
    │   ├── .directory
    │   ├── file2.py
    │   └── .sub_hidden2
    └── hidden_file.py
```


Now if we run this code on the above-mentioned directory we'll get the following `file_index`

```
[{'path': '/home/mzfr/dev/Test_folder/folder1', 'name': 'file.py'}, {'path': '/home/mzfr/dev/Test_folder/folder1', 'name': '.sub_hidden1'}, {'path': '/home/mzfr/dev/Test_folder/.hidden/folder2', 'name': '.sub_hidden2'}, {'path': '/home/mzfr/dev/Test_folder/.hidden/folder2', 'name': 'file2.py'}]
```

I think this is what we want.

Rechi:
The current PR title tells that everything in `.hidden` directory is ignored too, but the code doesn't ignore it.

kodi_addon_checker/handle_files.py
```diff
@@ -46,10 +46,14 @@ def create_file_index(path: str):
     """
     file_index = []
     for root, folders, files in os.walk(path, topdown=True):
+        if ".directory" in files:
```
Rechi: Why only skip directories named `.directory`?

mzfr: You mean to say we should ignore all the hidden files in the root directory?

Rechi: What is special about a folder named `.directory` that it should be ignored, but e.g. not a folder named `.abc`?
The current code does ignore `.directory` folders everywhere, not only in the root directory.
The current PR title is "Ignore hidden files and folders", which means ignore all files and folder for me.

mzfr:
@Razzeee Is there anything other changes required in this?

kodi_addon_checker/handle_files.py
Rechi: @Razzeee @mzfr I still don't get the reason to exclude a file named `.directory`. What's this special file that it shouldn't be listed in the file index?

mzfr: This file is basically used by file managers to keep track of the direcotries and also used for the preview of the direcotries.

Why I decided to ignore them? Since these files might be present in the folders(root) of addon but they don't consist any data that can be used by an addon so I simply decided to ignore them.

Rechi: So it's like `.DS_Store` and `Thumbs.db` on other OSes?

Still the checker shouldn't ignore the file, as  the checker now doesn't warn about those files submitted in an add-on submission.

PR #132, opened 2018-10-15
Added: ignore hidden directories and report an hidden top-directory

cybcafe:
#123 

Now as a matter of style: startswith or != ?

Also what to put in the report ?


cybcafe:
Checking with startswith on dirs excludes a lot of valid pathnames. Should have gone with a bigger conditional. Anyway I close this because there seem to be another PR ongoing for this issue.

PR #140, opened 2018-11-10
Parially revert change that made some tags not able to be empty

### LINKED ISSUE #133: Allow empty tags in addon.xml
Some one decided that empty tags like website,forum,email are not allowed. That's just wrong and needs to be fixed as they are perfectly valid.

Razzeee:
Closes #133

### DIFF
--- a/kodi_addon_checker/xml_schema/gotham_metadata.xsd
+++ b/kodi_addon_checker/xml_schema/gotham_metadata.xsd
@@ -8,14 +8,14 @@
         <xs:element name="description" type="translatedString" minOccurs="0" maxOccurs="unbounded"/>
         <xs:element name="disclaimer" type="translatedString" minOccurs="0" maxOccurs="unbounded"/>
         <xs:element name="platform" type="platformList" minOccurs="0" maxOccurs="1"/>
-        <xs:element name="language" type="nonEmptyString"  minOccurs="0" maxOccurs="1"/>
+        <xs:element name="language" type="xs:string"  minOccurs="0" maxOccurs="1"/>
         <xs:element name="license" type="nonEmptyString"  minOccurs="0" maxOccurs="1" />
-        <xs:element name="forum" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="forum" type="xs:string" minOccurs="0" maxOccurs="1"/>
         <xs:element name="website" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
         <xs:element name="source" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
-        <xs:element name="email" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="email" type="xs:string" minOccurs="0" maxOccurs="1"/>
         <xs:element name="broken" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
-        <xs:element name="news" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="news" type="nonEmptyStringCapped" minOccurs="0" maxOccurs="1"/>
         <xs:element name="reuselanguageinvoker" type="xs:boolean" minOccurs="0" maxOccurs="1"/>
         <xs:element name="assets" type="assetsList" minOccurs="0" maxOccurs="1"/>
       </xs:choice>
@@ -32,12 +32,16 @@
   <xs:simpleType name="nonEmptyString">
     <xs:restriction base="xs:string">
       <xs:pattern value="\s*([^\s]\s*)+"/>
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="nonEmptyStringCapped">
+    <xs:restriction base="nonEmptyString">
       <xs:maxLength value="1500"/>  
     </xs:restriction>
   </xs:simpleType>
   <xs:complexType name="translatedString">
     <xs:simpleContent>
-      <xs:extension base="nonEmptyString">
+      <xs:extension base="nonEmptyStringCapped">
         <xs:attribute name="lang" type="langIdentifier" use="required"/>
       </xs:extension>
     </xs:simpleContent>


kodi_addon_checker/xml_schema/gotham_metadata.xsd
```diff
@@ -8,12 +8,12 @@
         <xs:element name="description" type="translatedString" minOccurs="0" maxOccurs="unbounded"/>
         <xs:element name="disclaimer" type="translatedString" minOccurs="0" maxOccurs="unbounded"/>
         <xs:element name="platform" type="platformList" minOccurs="0" maxOccurs="1"/>
-        <xs:element name="language" type="nonEmptyString"  minOccurs="0" maxOccurs="1"/>
+        <xs:element name="language" type="xs:string"  minOccurs="0" maxOccurs="1"/>
         <xs:element name="license" type="nonEmptyString"  minOccurs="0" maxOccurs="1" />
-        <xs:element name="forum" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
-        <xs:element name="website" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="forum" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="website" type="xs:string" minOccurs="0" maxOccurs="1"/>
```
Rechi: Wiki doesn't say that website can be left blank.

Razzeee: Should be resolved

kodi_addon_checker/xml_schema/gotham_metadata.xsd
```diff
@@ -8,14 +8,14 @@
         <xs:element name="description" type="translatedString" minOccurs="0" maxOccurs="unbounded"/>
         <xs:element name="disclaimer" type="translatedString" minOccurs="0" maxOccurs="unbounded"/>
         <xs:element name="platform" type="platformList" minOccurs="0" maxOccurs="1"/>
-        <xs:element name="language" type="nonEmptyString"  minOccurs="0" maxOccurs="1"/>
+        <xs:element name="language" type="xs:string"  minOccurs="0" maxOccurs="1"/>
```
enen92: I'd leave language as a `nonEmptyString `. It is often a source of confusion between addon submitters and if included we should force users to define it. We can always edit the wiki

Razzeee: Can you please go ahead do that and edit the wiki? I will change the PR then.

Rechi:
I actually meant the following while talking about news
https://github.com/xbmc/addon-check/blob/9d9b8d89159cfa86b33a06962a0daf8cd81afad1/kodi_addon_checker/xml_schema/gotham_metadata.xsd#L35

Razzeee:
I did understand that. So you want me to make it mandatory to fill it, but not to cap it?

kodi_addon_checker/xml_schema/gotham_metadata.xsd
Rechi: So what's the difference to allow empty forum, email and news, but not language.
Either allow all to be empty, which in my opinion gives no additional info and bloats the addon.xml, or require all to be non empty or not added.

enen92: done

kodi_addon_checker/xml_schema/gotham_metadata.xsd
```diff
@@ -8,20 +8,20 @@
...
-        <xs:element name="source" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
-        <xs:element name="email" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
-        <xs:element name="broken" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="language" type="nonEmptyStringCapped"  minOccurs="0" maxOccurs="1"/>
+        <xs:element name="license" type="nonEmptyStringCapped"  minOccurs="0" maxOccurs="1" />
+        <xs:element name="forum" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="website" type="nonEmptyStringCapped" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="source" type="nonEmptyStringCapped" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="email" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="broken" type="nonEmptyStringCapped" minOccurs="0" maxOccurs="1"/>
```
Rechi: no max length mentioned for any of those elements in the wiki, they can be as long as wanted

kodi_addon_checker/xml_schema/gotham_metadata.xsd
MartijnKaijser: language must be allowed empty.

enen92: @Rechi I think language (unlike the others) is the only used in the Kodi GUI...likely ignored if empty though. I agree with you that empty tags only bloat the addon.xml but suddenly blocking user PRs is not a good alternative either. In the end, if Kodi does not complain it is up to the developer to have such a verbose manifest. We can warn during reviews if we spot such stuff

kodi_addon_checker/xml_schema/gotham_metadata.xsd
Razzeee: The initial trigger for this was https://github.com/xbmc/addon-check/pull/118 which was a limit for the news section. We need to change the wiki and at that point we can also think of limiting the other fields. But it's likely that we want shorter limits.

Until then I'm going to not cap them.

kodi_addon_checker/xml_schema/gotham_metadata.xsd
```diff
@@ -32,12 +32,18 @@
   <xs:simpleType name="nonEmptyString">
     <xs:restriction base="xs:string">
       <xs:pattern value="\s*([^\s]\s*)+"/>
-      <xs:maxLength value="1500"/>  
     </xs:restriction>
   </xs:simpleType>
-  <xs:complexType name="translatedString">
+  <xs:complexType name="nonEmptyStringCapped">
```
Rechi: no need for nonEmptyString and nonEmptyStringCapped, they are both the same

kodi_addon_checker/xml_schema/gotham_metadata.xsd
```diff
@@ -10,12 +10,12 @@
...
         <xs:element name="license" type="nonEmptyString"  minOccurs="0" maxOccurs="1" />
-        <xs:element name="forum" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="forum" type="xs:string" minOccurs="0" maxOccurs="1"/>
         <xs:element name="website" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
         <xs:element name="source" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
-        <xs:element name="email" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="email" type="xs:string" minOccurs="0" maxOccurs="1"/>
         <xs:element name="broken" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
-        <xs:element name="news" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="news" type="nonEmptyStringCapped" minOccurs="0" maxOccurs="1"/>
```
Rechi: wiki doesn't mention a limit

Razzeee:
Okay, another update. Can we agree, to do it like this and change the wiki?

Rechi:
If we change the wiki, I still not get why we would then require non empty strings for license, source and website, but not for email, forum, news and language.

I thought this whole PR is to keep the wiki as is, because otherwise the wiki could be changed for all elements and this PR wouldn't be needed.

MartijnKaijser:
What is current not according to the wiki?

Rechi:
With the current state of this PR, max length for
- description
- disclaimer
- news
- summary

Razzeee:
Current state of PR only max lengths:
* news

But I think the non empty stuff (no empty tags) for these are also not documented:
* license
* website
* source
* broken
* news (again)
* addon name
* icon
* fanart
* screenshot
* clearlogo
* banner

Razzeee:
I would like to merge this as is any more comments?

Rechi:
As already mentioned a few times in this PR, the wiki doesn't mention a char limit for news.

Razzeee:
Yes, thats what I'm going to change as soon as we merge this.

PR #139, opened 2018-11-08
Add license header to all python files.

mzfr:
Fixes #134 
I have added the small license header in all the python files in the tools. 
Should I also add the headers to [test files](https://github.com/xbmc/addon-check/tree/master/tests) and [test addon](https://github.com/xbmc/addon-check/tree/master/script.test)?

Also in the header will it be `Copyright (C) 2018-2018 Team Kodi` or `Copyright (C) 2005-2018 Team Kodi`?

PR #142, opened 2018-11-12
add reverse dependencies check

### LINKED ISSUE #124: Add check for orphaned add-on
Over time some add-ons might become broken/obsolete and will be removed. Some have a certain helper add-on which we then also forget to remove. Would be nice if we could detect this in some way if there are any. This can be done in the general scan on each branch and doesn't need to bother the regular submission report.

Rechi:
fixes #124

@MartijnKaijser is it correct to only warn for add-ons which have `script.module.*` as id?

### DIFF
--- a/kodi_addon_checker/addons/Repository.py
+++ b/kodi_addon_checker/addons/Repository.py
@@ -1,0 +1,40 @@
+"""
+    Copyright (C) 2018 Team Kodi
+    This file is part of Kodi - kodi.tv
+
+    SPDX-License-Identifier: GPL-3.0-only
+    See LICENSES/README.md for more information.
+"""
+
+from .Addon import Addon
+
+import gzip
+import requests
+import xml.etree.ElementTree as ET
+from io import BytesIO
+
+
+class Repository(object):
+    def __init__(self, version, path):
+        super(Repository, self).__init__()
+        self.version = version
+        self.path = path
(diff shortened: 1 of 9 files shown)


MartijnKaijser:
That should suffice for now I think.cant think of anything else

MartijnKaijser:
Shouldn't you add license header for the new files?

Rechi:
rebased on top of #139 and added license headers for the new files
needs another rebase after #139 gets merged

kodi_addon_checker/addons/Repository.py
```diff
@@ -0,0 +1,48 @@
...
+from io import BytesIO
+
+
+class Repository(object):
+    """docstring for Repository"""
+    def __init__(self, version, path):
+        super(Repository, self).__init__()
+        self.version = version
+        self.path = path
+        gz_file = requests.get(path, timeout=(10, 10)).content
```
enen92: maybe you can check the header first instead of assuming it is gzip

Rechi: This requests an explicit gz file and doesn't assume auto compression by the web server.
https://github.com/xbmc/addon-check/blob/b803624f41a8d0e84629063f3fc886d86ae0f458/kodi_addon_checker/check_addon.py#L29

PR #144, opened 2018-11-14
Add check for SPDX license identifier

### LINKED ISSUE #126: Use SPDX-License-Identifier for add-ons as well
For Kodi code we have switched to the SPDX-License-Identifier which makes it easier to identify what licenses are used. Perhaps this is a good idea as well for add-ons. Just as a suggestion for the dev but not make it mandatory.

mzfr:
Fixes #126 
Should this be reported as a `WARNING` or just `INFORMATION`?


Razzeee:
Not too excited about this, as it will generate a flood of messages either way.

mzfr:
Yes this will but another option is just say something like `SPDX IDENTIFIER MISSING` or one another thing I can think of, combine all the name of the files and print it like
```
SPDX Identifier missing from addon.py,addon1.py,addon2.py
```

mzfr:
@Razzeee I'll add a check for `license tag` in addon.xml file but that would still generate lots of warning.

Should we ignore files if the tag is present in the `addon.xml` assuming developer knows about the identifier thing?

MartijnKaijser:
License tag should already be mandatory. What this should do is see if doesn't match the SPDX format and suggest it. Same for the python files.

mzfr:
@Razzeee Is there any other changes required in this PR?

kodi_addon_checker/check_files.py
```diff
@@ -131,3 +132,33 @@ def check_file_permission(report: Report, file_index: list):
...
+                              relative_path(str(file))))
+
+
+def check_license_tag(report: Report, parsed_xml):
+    """Check license tag in addon xml file
+        :parsed_xml: parsed tree for xml file
+    """
+
+    license_text = parsed_xml.find("extension/license").text
+    spdx_string = "GNU General Public License v2.0 or later"
```
Rechi: This forces every add-on to be `GPLv2+`.
I don't think that was your intention.

mzfr: @Rechi There's a problem with the identification of Licenses, lot's of addon have `GNU GENERAL PUBLIC LICENSE Version 3, 29 June 2007`  this type of text in their `license` tag. Now should I follow the identifiers in https://spdx.org/licenses/, doing that will generate an error on the type I mentioned above or should I just check that the license tag is non-empty, but doing this will pass all types of Licenses including any random string in the license tag.

One solution could be that we create a dictionary of all the license that we need and then use it.

Rechi: As I already mentioned at another PR, first require the licence tag to be listed exactly once and have a non empty string as value.
Also there is no reason to disallow non standard licences.

Razzeee:
Do we really want these two checks? 

a) Not looking forward to having to have a license string in every file
b) Shouldn't be forcing a GNU license, like rechi correctly mentioned

PR #149, opened 2018-12-10
check presence of extensions in addon dependencies

### LINKED ISSUE #148: Addons are being validated despite not having xbmc.python
See: https://github.com/xbmc/repo-scripts/pull/947/files#r239163671

mzfr:
Fixes #148

### DIFF
--- a/kodi_addon_checker/check_dependencies.py
+++ b/kodi_addon_checker/check_dependencies.py
@@ -27,6 +27,26 @@ common_ignore_deps = ['xbmc.metadata.scraper.albums', 'xbmc.metadata.scraper.mov
                       'kodi.resource.font', 'kodi.inputstream', 'kodi.vfs', 'kodi.imagedecoder', 'xbmc.addon',
                       'xbmc.gui', 'xbmc.json', 'xbmc.metadata', 'xbmc.python', 'script.module.pil']
 
+extensions = {"kodi.gameclient": "kodi.binary.instance.game",
+              "xbmc.gui.skin": "xbmc.gui",
+              "kodi.vfs": "kodi.binary.instance.vfs",
+              "xbmc.metadata.scraper.albums": "xbmc.metadata",
+              "xbmc.metadata.scraper.artists": "xbmc.metadata",
+              "xbmc.metadata.scraper.library": "xbmc.metadata",
+              "xbmc.metadata.scraper.movies": "xbmc.metadata",
+              "xbmc.metadata.scraper.musicvideos": "xbmc.metadata",
+              "xbmc.metadata.scraper.tvshows": "xbmc.metadata",
+              "xbmc.pvrclient": "kodi.binary.instance.pvr",
+              "xbmc.python.library": "xbmc.python",
+              "xbmc.python.lyrics": "xbmc.python",
+              "xbmc.python.module": "xbmc.python",
+              "xbmc.python.pluginsource": "xbmc.python",
+              "xbmc.python.script": "xbmc.python",
+              "xbmc.python.weather": "xbmc.python",
+              "xbmc.ui.screensaver": "xbmc.python",
+              "xbmc.webinterface": "xbmc.webinterface",
+              }
+
 VERSION_ATTRB = {'xbmc.python': {'gotham': '2.14.0', 'helix': '2.19.0', 'isengard': '2.20.0',
                                  'jarvis': '2.24.0', 'krypton': '2.25.0', 'leia': '2.26.0'}}
 
@@ -73,6 +93,8 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
                                       .format(branch_name, dependency.id, version)))
             except KeyError:
                 LOGGER.warn("Misconfiguration in VERSION_ATTRB of check_dependencies")
+
+    _check_extensions(report, parsed_xml, addon)
 
 
 def check_reverse_dependencies(report: Report, addon: str, branch_name: str, all_repo_addons: dict):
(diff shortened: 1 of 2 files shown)


kodi_addon_checker/check_dependencies.py
```diff
@@ -56,6 +56,9 @@ def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, bran
                                    "version mismatch for optional dependencies: %s. Available: %s, Required: %s" %
                                    (dependency, deps_version, version))
 
+    if 'xbmc.python' not in deps:
```
Rechi: Not every add-on has to list `xbmc.python` as dependency, only python add-ons.

Rechi: Python is the only possible scripting language, but skins, languages, controllers, webinterfaces, ... don't use python. So they shouldn't (and also do not) import `xbmc.python`.

mzfr:
@Rechi Now we check for `xbmc.python` only when there is any python file present in the addon submitted. This way if an addon is of any other language then it won't look for xbmc.python dependencies

kodi_addon_checker/check_dependencies.py
```diff
@@ -31,7 +32,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, branch_name: str):
+def check_addon_dependencies(report: Report, repo_addons: dict, parsed_xml, branch_name: str, file_index):
```
Razzeee: Can you add the type?

Razzeee:
Should work fine for now. Checking just for py files might cause some errors in some obscure cases.

Rechi:
Checking for if py files exist somewhere in the add-on folder is the wrong approach.
The correct one is to check if any python extension point in addon.xml is listed.

mzfr:
@Rechi so you are saying instead of looking for a file. I should check the library of extension point?
Like
```
 <extension point='xbmc.python.pluginsource' library='resources/lib/routes.py'>
```

and then check if the library has `.py` extension file in the end?

Rechi:
Yes exactly like that, but xbmc.python.pluginsource isn't the only extension point which uses python.
I'm not sure if the .py file extension check is actually required. You have to look into Kodi code if it always assumes python or does some checks.

mzfr:
No, we are having `.py` extension check to actually see that whether it is a python addon or not. So even if there is no check in Kodi code we'll still need it so we can differentiate between other addon and python addon.

Rechi:
As Kodi doesn't require `.py` file extension for e.g. `xbmc.python.pluginsource`, the add-on checker shouldn't assume that file extension is always used.
If wanted you can add an additional check if the correct file extension is used for every extension point, but in a separate PR.

kodi_addon_checker/check_dependencies.py
```diff
@@ -27,6 +27,17 @@
                       'kodi.resource.font', 'kodi.inputstream', 'kodi.vfs', 'kodi.imagedecoder', 'xbmc.addon',
                       'xbmc.gui', 'xbmc.json', 'xbmc.metadata', 'xbmc.python', 'script.module.pil']
 
+extensions = {"kodi.gameclient": "kodi.binary.instance.game", "kodi.vfs": "kodi.binary.instance.vfs",
+              "xbmc.gui.skin": "xbmc.gui", "xbmc.metadata.scraper.albums": "xbmc.metadata",
```
Rechi: Moving every key value pair into a seperate line, would make it more easily to extend.

kodi_addon_checker/check_dependencies.py
```diff
@@ -27,6 +27,17 @@
...
+              "xbmc.metadata.scraper.artists": "xbmc.metadata", "xbmc.metadata.scraper.library": "xbmc.metadata",
+              "xbmc.metadata.scraper.movies": "xbmc.metadata", "xbmc.metadata.scraper.musicvideos": "xbmc.metadata",
+              "xbmc.metadata.scraper.tvshows": "xbmc.metadata", "xbmc.pvrclient": "kodi.binary.instance.pvr",
+              "xbmc.python.library": "xbmc.python", "xbmc.python.lyrics": "xbmc.python",
+              "xbmc.python.module": "xbmc.python", "xbmc.python.pluginsource": "xbmc.python",
+              "xbmc.python.script": "xbmc.python", "xbmc.python.weather": "xbmc.python",
+              "xbmc.ui.screensaver": "xbmc.python", "xbmc.webinterface": "xbmc.webinterface",
```
Rechi: `xbmc.ui.screensaver` doesn't always require `xbmc.python`.

A screensaver can either be a binary add-on or a python add-on.

kodi_addon_checker/check_dependencies.py
```diff
@@ -122,3 +135,19 @@ def _get_ignore_list(branch_name):
 
     else:
         return common_ignore_deps
+
+
+def check_extensions(report: Report, parsed_xml, addon):
```
Rechi: As this function is not supposed to be called from outside of this file, mark it pseudo private.
```suggestion
def _check_extensions(report: Report, parsed_xml, addon):
```

kodi_addon_checker/check_dependencies.py
```diff
@@ -122,3 +135,19 @@ def _get_ignore_list(branch_name):
...
+
+      :addon: class kodi_addon_checker.addons.Addon.Addon
+
+    """
+    deps = [dependency.id for dependency in addon.dependencies]
+
+    for extension in parsed_xml.findall("extension"):
+        point = extension.get("point")
+        if point in extensions and extensions[point] not in deps:
+            report.add(Record(PROBLEM, "Required dependency {} is missing from the addons"
```
Rechi: This exact message is already used to point out that an dependency listed in addon.xml is not available the official Kodi repo.
Here it is used to point out that a dependency has to be added to addon.xml.

So one of those has to be repharsed to not cause confussion.

kodi_addon_checker/check_dependencies.py
mzfr: So should I remove it from the list?

Rechi: If you do not plan to implement it correct, remove it.

Rechi:
Feature list update in README.md is missing.

PR #150, opened 2018-12-10
Add check for print statement

### LINKED ISSUE #19: Search for print statements and report them

mzfr:
Fixes #19


kodi_addon_checker/check_files.py
```diff
@@ -131,3 +131,19 @@ def check_file_permission(report: Report, file_index: list):
...
+       or not
+        :file_index: list having names and path of all the files present in addon
+    """
+
+    for file in file_index:
+        if os.path.splitext(file["name"])[1] == '.py':
+            file = os.path.join(file["path"], file["name"])
+            with open(file, 'r') as f:
+                for ind, line in enumerate(f, 1):
+                    if 'print(' in line or 'print (' in line:
```
Rechi: This could be inside a block comment or a different method e.g. `someClass.print(...)`.

mzfr:
@Rechi I have used regex to find the print statement and I have tested this with the following content in a file.

```
"""
This is docstings and here is the print("testing")
"""

print("testing separately")

someclasss.print("with some function") # comment with print("in a comment")
# comment with print("in a comment separately")
```

and the output was `print("testing separately")` only

Rechi:
The following isn't detected.
```
def printThat(that):
    print(that)
```

Rechi:
`if True: print('ABC')` isn't detected.

mzfr:
@Rechi That would just make it's difficult to catch because if I remove the `^` from the regex pattern then it would start counting all the print statements including the comments etc

Rechi:
Doing the search for print with reg-ex is the wrong approach. You have to use some parsers which correctly understands the python syntax.

mzfr:
@Rechi I did try with AST but again that didn't covered all the cases

mzfr:
@Rechi I did try with AST but again that didn't covered all the cases

mzfr:
@Rechi So I tried again with the ast and I figured it out. So Now it should report any print statement ignoring the docstrings and comments.

kodi_addon_checker/check_addon.py
```diff
@@ -59,6 +59,8 @@ def start(addon_path, args, all_repo_addons, config=None):
 
             check_dependencies.check_reverse_dependencies(addon_report, addon_id, args.branch, all_repo_addons)
 
+            check_files.check_print_statement(addon_report, file_index)
```
Rechi: This check should be either run just before or just after `check_py3_compatibility`.

kodi_addon_checker/check_files.py
```diff
@@ -139,3 +140,27 @@ def check_file_permission(report: Report, file_index: list):
...
+                source = f.read()
+
+            try:
+                for node in ast.walk(ast.parse(source)):
+                    if (isinstance(node, ast.Expr) and
+                            isinstance(node.value, ast.Call) and
+                            isinstance(node.value.func, ast.Name)):
+
+                        if node.value.func.id == "print":
+                            report.add(Record(WARNING, "%s has a print statement" % relative_path(str(file))))
```
Rechi: Also report the line (`node.lineno`).

kodi_addon_checker/check_files.py
```diff
@@ -139,3 +140,27 @@ def check_file_permission(report: Report, file_index: list):
...
+            except SyntaxError:
```
Rechi: SyntaxError gives a description and line number of the error, so that should reported.
```suggestion
            except SyntaxError as e:
```

kodi_addon_checker/check_files.py
```diff
@@ -139,3 +140,27 @@ def check_file_permission(report: Report, file_index: list):
...
+                report.add(Record(PROBLEM, "Invalid code present in file %s" % (relative_path(file))))
```
Rechi:
```suggestion
                report.add(Record(PROBLEM, "%s failed to parse with %s" % (relative_path(str(file)), e)))
```

kodi_addon_checker/check_files.py
```diff
@@ -139,3 +140,28 @@ def check_file_permission(report: Report, file_index: list):
...
+def check_print_statement(report: Report, file_index: list):
+    """Check whether any addon files have a print statement in them
+       or not
+        :file_index: list having names and path of all the files present in addon
+    """
+    for file in file_index:
+        if os.path.splitext(file["name"])[1] == '.py':
+            file = os.path.join(file["path"], file["name"])
+
+            with open(file) as f:
```
Rechi: I tested it on windows and this needs an explicit `encoding="UTF-8"`. That's the default for Linux so you may have not noticed any issues.

kodi_addon_checker/check_files.py
```diff
@@ -139,3 +140,28 @@ def check_file_permission(report: Report, file_index: list):
...
+def check_print_statement(report: Report, file_index: list):
+    """Check whether any addon files have a print statement in them
+       or not
+        :file_index: list having names and path of all the files present in addon
+    """
+    for file in file_index:
+        if os.path.splitext(file["name"])[1] == '.py':
+            file = os.path.join(file["path"], file["name"])
+
+            with open(file, 'r', encoding="utf-8") as f:
```
Rechi: `open` may raise a `UnicodeDecodeError`.

kodi_addon_checker/check_files.py
```diff
@@ -139,3 +140,28 @@ def check_file_permission(report: Report, file_index: list):
...
+
+            try:
+                for node in ast.walk(ast.parse(source)):
+                    if (isinstance(node, ast.Expr) and
+                            isinstance(node.value, ast.Call) and
+                            isinstance(node.value.func, ast.Name)):
+
+                        if node.value.func.id == "print":
+                            report.add(Record(WARNING, "%s has a print statement on line %s" %
+                                              (relative_path(str(file)), node.line)))
```
Rechi: `node.line` is wrong, is may not exist. Correct is `node.lineno`.

Rechi:
IMO print statements shouldn't be reported for `script.module` add-ons, as most of those are not specifically coded for Kodi.

@enen92 @Razzeee what's your opinion?

enen92:
@Rechi agreed. Also not for any file containing `test` or within a `test` folder.

Rechi:
Yes test should be ignored too, but not explicit by this check. `handle_files.create_file_index` shouldn't return test files.

mzfr:
@Rechi 	Why are we ignoring the test files? I mean the @MartijnKaijser 	[comment](https://github.com/xbmc/addon-check/issues/19#issuecomment-351807896) about people using print statement in unit test make sense.

Rechi:
Your second sentence already answers the question. In unit tests one can't use `xbmc.log(...)`, because they aren't from inside Kodi. If tests aren't ignored you could get lot of print used reports for those.

PR #151, opened 2018-12-30
require icon/fanart be explicitly declared in addon.xml assets

rmrector:
Implicit fanart.jpg and icon.png in the root of the add-on directory was superseded by the explicit `assets.fanart` and `assets.icon` declarations in Kodi Krypton 17 with xbmc/xbmc#10060. There is a bit of awkward backward compatibility in Kodi core that extends to add-on lists in skins with the implicit artwork, so I'd like to generally require "assets" be used for further submissions so this back-compat can be removed at some point.

Noted on the wiki pages [Add-on structure#icon.png](https://kodi.wiki/view/Add-on_structure#icon.png) and [Addon.xml#&lt;assets&gt;](https://kodi.wiki/view/Addon.xml#.3Cassets.3E).

It should also be noted in the forums.

### DIFF
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -64,7 +64,7 @@ def start(addon_path, args, all_repo_addons, config=None):
 
             check_files.check_for_invalid_json_files(addon_report, file_index)
 
-            check_artwork.check_artwork(addon_report, addon_path, parsed_xml, file_index)
+            check_artwork.check_artwork(addon_report, addon_path, parsed_xml, file_index, args.branch)
 
             max_entrypoint_count = config.configs.get(
                 "max_entrypoint_count", 15)
--- a/kodi_addon_checker/check_artwork.py
+++ b/kodi_addon_checker/check_artwork.py
@@ -17,7 +17,7 @@ from .record import PROBLEM, Record, WARNING, INFORMATION
 LOGGER = logging.getLogger(__name__)
 
 
-def check_artwork(report: Report, addon_path: str, parsed_xml, file_index: list):
+def check_artwork(report: Report, addon_path: str, parsed_xml, file_index: list, branch_name: str):
     """Checks for icon/fanart/screenshot
         :addon_path: path to the folder having addon files
         :parsed_xml: xml file i.e addon.xml
@@ -25,7 +25,7 @@ def check_artwork(report: Report, addon_path: str, parsed_xml, file_index: list)
     """
     art_type = ['icon', 'fanart', 'screenshot']
     for image_type in art_type:
-        _check_image_type(report, image_type, parsed_xml, addon_path)
+        _check_image_type(report, image_type, parsed_xml, addon_path, branch_name)
 
     for file in file_index:
         if re.match(r"(?!fanart\.jpg|icon\.png).*\.(png|jpg|jpeg|gif)$", file["name"]) is not None:
@@ -37,7 +37,7 @@ def check_artwork(report: Report, addon_path: str, parsed_xml, file_index: list)
                     Record(PROBLEM, "Could not open image, is the file corrupted ? %s" % relative_path(image_path)))
 
 
-def _check_image_type(report: Report, image_type: str, parsed_xml, addon_path: str):
+def _check_image_type(report: Report, image_type: str, parsed_xml, addon_path: str, branch_name: str):
     """Check for whether the given image type exists or not if they do """
 
     fallback, images = _assests(image_type, parsed_xml, addon_path)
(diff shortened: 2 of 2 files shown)


kodi_addon_checker/check_artwork.py
```diff
@@ -48,6 +48,8 @@ def _check_image_type(report: Report, image_type: str, parsed_xml, addon_path: s
 
             if os.path.isfile(filepath):
                 report.add(Record(INFORMATION, "Image %s exists" % image_type))
+                if fallback:
+                    report.add(Record(PROBLEM, "Image %s should be explicitly declared in <assets>." % image_type))
```
Rechi: This may only be reported for krypton and higher.

Razzeee:
Fine for me.

CC @MartijnKaijser @ronie

kodi_addon_checker/check_artwork.py
```diff
@@ -48,6 +48,9 @@ def _check_image_type(report: Report, image_type: str, parsed_xml, addon_path: s
 
             if os.path.isfile(filepath):
                 report.add(Record(INFORMATION, "Image %s exists" % image_type))
+                if fallback and branch_name not in ['gotham', 'helix', 'isengard', 'jarvis']:
+                    report.add(Record(PROBLEM,
+                        "Image %s should be explicitly declared in addon.xml <assets>." % image_type))
```
Rechi: > `./kodi_addon_checker/check_artwork.py:53:25: E128 continuation line under-indented for visual indent`

enen92:
Agreed too. We should drop a lot of deprecated methods and arguments in v19 as well: thumbnailimage, iconImage, etc

ronie:
since most addon devs will need to update their addon to py3 anyway for v19,
now would be the perfect time to add some more required changes.

apart from the ones mentioned above, we'd also could drop support for xml based language files,
language folder names (English vs. resource.language.en_GB) and language attributes in addon.xml (lang="en" vs lang="en_GB")

besides updating the wiki with this info, all required changes need to be published on the forum as well.

Razzeee:
I really like that idea

Rechi:
> we'd also could drop support for xml based language files

That's already done for all versions (gotham - current).
https://github.com/xbmc/addon-check/blob/b803624f41a8d0e84629063f3fc886d86ae0f458/kodi_addon_checker/check_addon.py#L79

> language folder names (English vs. resource.language.en_GB)

That's already done for isengard - current.
https://github.com/xbmc/addon-check/blob/b803624f41a8d0e84629063f3fc886d86ae0f458/kodi_addon_checker/check_addon.py#L81-L82

Both checks result in errors (not warnings).

**PS**: Please move this discussion somewhere else as it is off topic to this PR.

kodi_addon_checker/check_artwork.py
rmrector: obnoxious linter rule satisfied

rmrector:
@Rechi So I'm torn on only reporting for Krypton and above branches. Maybe we should usually expect add-ons to target the current Kodi version, even if the add-on supports an older Kodi version. For example, image resource add-ons were introduced in Jarvis, so plenty are still there (and a few get updates), but still work the same in Krypton, Leia, and likely into the future.

Rechi:
Are you aware that new add-ons are only allowed for krypton and leia (https://kodi.wiki/view/Submitting_Add-ons#Allowed_submissions)? For lower versions only updates are allowed.

Lower versions don't understand the assets tag. If people are forced to add that tag, they may choose non default paths (`icon.png`, `fanart.jpg`). That would break getting icon and fanart in Kodi/XBMC versions Gotham - Jarvis.

rmrector:
Yes. But I still think updates to old repos should aim for compatibility with the current version of Kodi when forward compatibility is feasible - the repos rely on many add-on interfaces being forward compatible.

They also don't hurt anything for lower (unsupported, fewer users) versions, but do and will help current and new versions. And it can be clearly documented that for compatibility with older versions of Kodi they can be kept in the same place, like the initial PR does (xbmc/xbmc#10060).

Rechi:
> They also don't hurt anything for lower versions.

Yes as long as the default paths are used for that versions. So if you want to require assets tag for every version, there has to be a check that default paths are used for add-ons targeting Gotham - Jarvis.

rmrector:
Nope, not interested in doing more work for old unsupported versions of Kodi. I'll leave it the way it is.

rmrector:
@ronie Your name is on a lot of the important forum posts for this kind of thing, do you want to post this one?

Noted in the wiki, exact sections linked in the OP.

kodi_addon_checker/check_addon.py
```diff
@@ -61,7 +61,7 @@ def start(addon_path, branch_name, all_repo_addons, pr, config=None):
 
             check_files.check_for_invalid_json_files(addon_report, file_index)
 
-            check_artwork.check_artwork(addon_report, addon_path, parsed_xml, file_index)
+            check_artwork.check_artwork(addon_report, addon_path, parsed_xml, file_index, branch_name)
```
Rechi: This needs a rebase on top of master and the following change.
```suggestion
            check_artwork.check_artwork(addon_report, addon_path, parsed_xml, file_index, args.branch)
```

PR #153, opened 2019-01-24
2to3: split fixers into required and suggested

Rechi:
This will report python syntax incompatibilities as error for all versions. The new syntax of those checks was backported to Python 2.6 and therefore is compatible with Gotham and newer.


mzfr:
@Rechi there is something wrong with the commits, you have added the ` Revert "Ignore hidden files and folders" ` commit in this PR also.

Rechi:
@enen92 @Razzeee ping

PR #154, opened 2019-01-24
add url (forum, source, website) check

Rechi:


kodi_addon_checker/check_url.py
```diff
@@ -0,0 +1,53 @@
...
+    sources += parsed_xml.findall("./extension[@point='xbmc.addon.metadata']/forum")
+
+    sources += parsed_xml.findall("./extension[@point='kodi.addon.metadata']/source")
+    sources += parsed_xml.findall("./extension[@point='xbmc.addon.metadata']/source")
+
+    sources += parsed_xml.findall("./extension[@point='kodi.addon.metadata']/website")
+    sources += parsed_xml.findall("./extension[@point='xbmc.addon.metadata']/website")
+
+    for source in sources:
+        if source.text is None or source.text == "":
```
enen92: If not source.text

Rechi:
As https://kodi.tv/addons and kodi.wiki lists forum, source and website from addon.xml those should be valid urls.
@enen92 if wanted I can change the errors to warnings, so that the url validation never fails a check.

Rechi:
The reason for implementing this, was that someone searched the upstream repo for an add-on, but the url in the addon.xml was wrong and pointed to a 404 page. You may want check the logs about how many invalid urls are listed in the addon.xmls.

Razzeee:
Okay, not very surprising. Code read and previous interactions sounded like we're afraid to link to problematic websites.

anxdpanic:
Wouldn't using requests.get() here instead of requests.head() leave this open to abuse? ie. multiple PRs with forum/website/source all point to > 1GB files

Sorry for creating noise if I missed something here.

Rechi:
@anxdpanic Thanks for the hint. Now a the check runs also quicker.

PR #157, opened 2019-02-02
add argument --allow-folder-id-mismatch to report addon_xml_matches_folder as INFO

anxdpanic:
w/ arg:
```INFO: Addon id and folder name does not match. Ensure folder name is <addon_id> when submitting a PR to Kodi's official repository.```

w/o arg:
```ERROR: Addon id and folder name does not match.```

### DIFF
--- a/kodi_addon_checker/__main__.py
+++ b/kodi_addon_checker/__main__.py
@@ -38,7 +38,7 @@ def dir_type(dir_path):
     return os.path.abspath(dir_path)
 
 
-def check_artifact(artifact_path, args, branch_name, all_repo_addons, pr):
+def check_artifact(artifact_path, args, branch_name, all_repo_addons):
     """
     Check given artifact and return its report. The artifact can be either an add-on or a repository.
     :param artifact_path: the path of add-on or repo
@@ -52,9 +52,9 @@ def check_artifact(artifact_path, args, branch_name, all_repo_addons, pr):
     config = Config(artifact_path, args)
     ConfigManager.process_config(config)
     if os.path.isfile(os.path.join(artifact_path, "addon.xml")):
-        return check_addon.start(artifact_path, branch_name, all_repo_addons, pr, config)
+        return check_addon.start(artifact_path, branch_name, all_repo_addons, args, config)
     else:
-        return check_repo(artifact_path, branch_name, all_repo_addons, pr, config)
+        return check_repo(artifact_path, branch_name, all_repo_addons, args, config)
 
 
 def main():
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: kodi_addon_checker/__main__.py [hunk 1]
```diff
@@ -52,7 +52,7 @@ def check_artifact(artifact_path, args, branch_name, all_repo_addons, pr):
     config = Config(artifact_path, args)
     ConfigManager.process_config(config)
     if os.path.isfile(os.path.join(artifact_path, "addon.xml")):
-        return check_addon.start(artifact_path, branch_name, all_repo_addons, pr, config)
+        return check_addon.start(artifact_path, branch_name, all_repo_addons, pr, not pr, config)
```
Razzeee: Not a fan of `sa` as the combination of `pr` and `sa` are not all handled (or available).
Any reason your not using `pr` directly?

Razzeee:
Should be fine for the kodi repo usecases, but probably won't work for the ci of single repos.

anxdpanic:
Should I change this to have it's own argument ```--skip-folder-matches-id``` and change the warning to info?

Rechi:
Yes it should be changed to have it's own argument.
The name `--skip-folder-matches-id` would make me think that actually executing the `addon_xml_matches_folder` function is completely skipped, not that reports a warning or info instead of an error. But that's nothing I care much about.

anxdpanic:
Any feelings on warning vs info? I went with warning here since it's a must when submitting.

Rechi:
The following line needs an update too.
https://github.com/xbmc/addon-check/blob/ad8aa4960a0fefdb2f4661ff9fd7f2e90427e445/tests/test_check_addon.py#L26

anxdpanic:
Updated test_check_addon, required adding an ```Args``` class as well

kodi_addon_checker/__main__.py
```diff
@@ -74,6 +74,7 @@ def main():
     parser.add_argument("--branch", choices=choice, required=True,
                         help="Target branch name where the checker will resolve dependencies")
     parser.add_argument("--PR", help="Tell if tool is to run on a pull requests or not", action='store_true')
+    parser.add_argument("--allow-folder-id-mismatch", help="Allow the addon's folder name and id to mismatch", action="store_true")
```
Rechi: > ./kodi_addon_checker/__main__.py:77:121: E501 line too long (131 > 120 characters)

Razzeee:
The new argument needs to be documented in the readme

anxdpanic:
Added to the readme and shifted the description alignment accordingly

Razzeee:
thank you!

PR #158, opened 2019-02-06
change --PR to --higher-version-required

anxdpanic:
I wasn't sure if a complete rename was wanted, so I left --PR usable so the transition to the new argument doesn't need to be immediate with the next release.


Rechi:
This is exactly what I had in mind. Thanks for also thinking about our current scripts and keeping the `--PR` argument functional for now.

Only thing missing is updating the parameter in some (example) configuration files, namely `.travis-addon.yml.example`, `.travis-repo.yml.example` and `.vscode/launch.json`.

anxdpanic:
Updated those files now. ```.travis-repo.yml.example``` has a line that is too long, will look at adding and testing newline addition to the script if required

Rechi:
`pycodestyle` doesn't check that file so all fine.

Razzeee:
To be honest, I would prefere configuration by use-case not configuration by feature of the addon checker.

So having some prepared grouped features that get triggered by a arg would be best in my mind.

`addon checker --mode=PR`

anxdpanic:
That seems like the opposite direction of #157
I think that would require 4 modes at least, ADDON being split into local filesystem and ci to account for stuff like gitignored files.

Rechi:
@Razzeee what would you enable or disable in each of those modes?

The number of all combinations of `--higher-version-required` and `--allow-folder-id-mismatch` is only 4. I don't see how having 3 different modes would be simper, as you then don't have the obvious argument names anymore.
I do not expect that we add any further command line arguments for disabling features.

Razzeee:
Yes, we might not expect it, but still we might have to do that at one point.

Far more important is that those arguments need input at least the `whitelist/blacklist` ones. So saying it's just four isn't really fair. So either everybody will just copy the default ones we provide or every addon author needs to set them up on their own.

PR #159, opened 2019-02-06
Configurable file exclusions

anxdpanic:
To address #156 , and possibly applies to #123

Allow for environment flexibility
```
--exclude-file              files/folders to exclude from permissions and file extension checks
--exclude-file-ext          file extensions to exclude from permissions and file extension checks
--whitelist-file            files/folders to exclude from file extension checks
--whitelist-file-ext        file extensions to exclude from file extension checks
```

PR #161, opened 2019-02-09
fix test_check_files on Windows

anxdpanic:


anxdpanic:
Separated from #159 so I don't forget if it's not accepted

PR #165, opened 2019-02-14
add check for circular dependencies

anxdpanic:


anxdpanic:
Wasn't sure how to approach a test for this at first, working on that now though.

anxdpanic:
Test is added now

kodi_addon_checker/check_dependencies.py
```diff
@@ -132,6 +132,32 @@ def check_reverse_dependencies(report: Report, addon: str, branch_name: str, all
...
+        :all_repo_addons: dictionary return by all_repo_addon() function
+        :parsed_xml: parsed addon.xml file
+        :branch_name: name of the kodi branch/version
+    """
+    addon = Addon(parsed_xml)
+    ignore = _get_ignore_list(branch_name)
+
+    dependencies = []
+    for dependency in addon.dependencies:
+        if dependency.id in ignore or dependency.id in VERSION_ATTRB:
```
Rechi: `VERSION_ATTRB` doesn't tell anything about repo availability. Optional dependencies have to be excluded.
```suggestion
        if dependency.id in ignore or dependency.optional:
```

kodi_addon_checker/check_dependencies.py
```diff
@@ -132,6 +132,32 @@ def check_reverse_dependencies(report: Report, addon: str, branch_name: str, all
...
+            continue
+        dependencies.append(dependency)
+
+    circular_dependencies = []
+    for depend in dependencies:
```
Rechi: This check has to be recursive.

tests/test_check_dependencies.py
```diff
@@ -0,0 +1,39 @@
...
+    def setUp(self):
+        load_plugins()
+        ReportManager.enable(["array"])
+        self.report = Report("")
+        self.branch = 'krypton'
+
+    def test_check_circular_dependency(self):
+        self.path = join(HERE, 'test_data', 'Circular depend')
+        addon_xml = join(self.path, "addon.xml")
+        addons_xml_gzip = join(self.path, "addons.xml.gz")
```
Rechi: use a plain xml file

tests/test_data/Circular depend/addon.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version='1.0' encoding='utf-8'?>
```
Rechi: remove or replace the space from the file path

kodi_addon_checker/check_dependencies.py
anxdpanic: We should check `VERSION_ATTRB` entries for circular?
I would think you'd still want to check if an optional dependency is circular because that would be an issue if the user has it installed?

anxdpanic:
I have made the required changes, other than the optional/attrib until further discussion

I added some temp logging to show the results against a repo,
https://pastebin.com/10Gkw9Jx

kodi_addon_checker/check_dependencies.py
Rechi: `VERSION_ATTRB` is just to give hints on updating the imported version of `xbmc.python` (and in future maybe others). It doesn't define if an addon exists or not. That's defined by `_get_ignore_list(...)` and the add-ons available in the repo.

If you want to check optional dependencies too, I'm fine with that. Those wouldn't have the ordering during add-on installation (if it is correctly implemented in Kodi).

anxdpanic: I completely missed `xbmc.python` in the ignore list, which is why I used the in `VERSION_ATTRB`.
Will remove that.

anxdpanic: Removed `in VERSION_ATTRB` now

anxdpanic:
rebased and resolved merge conflicts and pylint assertions

kodi_addon_checker/addons/Repository.py
```diff
@@ -16,13 +16,19 @@
 
 
 class Repository():
-    def __init__(self, version, path):
+    def __init__(self, version, path, local_xml=False):
```
Razzeee: can you add docstrings here, as your adding a new param (it should be documented that it's only use is testing)

anxdpanic: Done, I also added simple descriptive doc strings to the unit test

kodi_addon_checker/check_dependencies.py
```diff
@@ -139,6 +139,59 @@ def check_reverse_dependencies(report: Report, addon: str, branch_name: str, all
...
+        if dependencies[i].id not in dependency_tree:
+            for _, repo in sorted(all_repo_addons.items()):
+                found_addon = repo.find(dependencies[i].id)
+                if found_addon:
+                    add_dependencies(found_addon)
+        i += 1
+
+    if dependency_tree:
+        circular_dependencies = []
+        dependencies = list(dependency_tree.keys())
```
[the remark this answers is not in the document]
anxdpanic: something like `all_depends` work?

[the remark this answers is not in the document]
Razzeee: sure, just not `dependencies` :D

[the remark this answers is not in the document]
anxdpanic: All set :)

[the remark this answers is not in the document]
anxdpanic: okay, so that causes too many locals, with pylint. will have to take another look.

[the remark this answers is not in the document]
anxdpanic: removed the variable, to satisfy the pylint assertion as well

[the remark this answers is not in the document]
Razzeee: works

Razzeee:
Would be great if @Rechi could sign this off too :)

anxdpanic:
I've removed the local_xml parameter, and moved dependency tree creation to it's own function with it's own test

Issue #173, opened 2019-03-08
Switching PEP style guideline check from pycodestyle to pylint.


LordGameleo:
> Well it's running on the ci https://travis-ci.org/xbmc/addon-check/jobs/494757150#L520
Maybe we have something misconfigured or we just switch to pylint.

_Originally posted by @Razzeee in https://github.com/xbmc/addon-check/issues/168#issuecomment-469486822_

LordGameleo:
@Razzeee Should I start working on this issue?

Razzeee:
I would recommend to wait until https://github.com/xbmc/addon-check/pull/172 is merged

mzfr:
@LordGameleo I'll complete the work on that PR by tonight.

LordGameleo:
I will submit it as early as possible

LordGameleo:
@Razzeee I am closing this issue and creating new issues which solves particular problem of pylint checks we disabled

PR #174, opened 2019-03-10
Switched from pycodestyle to pylint

LordGameleo:
#173 
**Note:** It will not pass the build. We will have to correct it. I have created new issue for it. #175

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,51 @@
+[MASTER]
+
+# Add files or directories to the blacklist.
+ignore=LICENSE,
+       MANIFEST.in,
+       PUBLISH.md,
+       README.md,
+       requirements.txt,
+       script.test,
+
+
+[MESSAGES CONTROL]
+
+# Types of check we want to disable
+# format
+# disable=stuff-that-we-actually-want-ignored,
+#        stuff-that-we-actually-want-ignored2,
+#
+#        stuff-thats-only-ignored-for-now-and-will-be-addressed-in-a-pr,
+#        stuff-thats-only-ignored-for-now-and-will-be-addressed-in-a-pr2
+disable=import-error,
+        no-name-in-module,
+
+        missing-docstring,
+        broad-except,
+        too-few-public-methods,
+        invalid-name,
+        attribute-defined-outside-init,
+        useless-object-inheritance,
+        no-else-return,
+        len-as-condition,
+        duplicate-code,
+        unused-import,
+        unused-variable,
+        useless-else-on-loop,
+        bad-continuation,
+        too-many-arguments,
+        logging-not-lazy,
+        unidiomatic-typecheck,
+        too-many-nested-blocks,
+        inconsistent-return-statements,
+        arguments-differ,
+        consider-using-in,
+        deprecated-method,
+        redefined-outer-name,
+
+
+[FORMAT]
+
+# Maximum number of characters on a single line.
+max-line-length=120
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,7 @@ matrix:
 # command to install dependencies
 install:
   - pip install -r requirements.txt
+  - pip install pylint
 
 before_script:
   - if [[ -n "$REPO" && -n "$BRANCH" ]]; then
@@ -94,5 +95,5 @@ script:
       py.test -v;
     fi
   - if [[ -z "$REPO" || -z "$BRANCH" ]]; then
-      pycodestyle . --max-line-length=120;
+      pylint *;
     fi


mzfr:
I think we should have a .pylintrc too. So we can configure pylint. Because pylint have lots of options of suggestions and stuff which might not be needed.

LordGameleo:
@mzfr I will add that in a while I am having my exams this week.

LordGameleo:
What I propose is:
- We can have multiple issues which focuses on solving these.
- We need to discuss which all checks we need according to kodi's style of code.
- For now I will just put all these checks in disabled form so our build passes and as we progress in future with these smaller issues which aims at one type of check at a time we will remove them from pylintrc file.
- It will create few first timers only/newcomers issues.

@Razzeee @Rechi @mzfr  What do you think about it?

LordGameleo:
I disabled these checks to get a 10 on 10 score for now.
```    
        broad-except,
        too-few-public-methods,
        parse-error,
        invalid-name,
        attribute-defined-outside-init,
        pointless-statement,
        useless-object-inheritance,
        no-else-return,
        no-name-in-module,
        len-as-condition,
        missing-docstring,
        bad-whitespace,
        duplicate-code,
        unused-import,
        unused-variable,
        useless-else-on-loop,
        bad-continuation,
        import-error,
        too-many-arguments,
        logging-not-lazy,
        undefined-variable,
        syntax-error,
        unidiomatic-typecheck,
        too-many-nested-blocks,
        inconsistent-return-statements,
        arguments-differ,
        consider-using-in,
        deprecated-method,
```

LordGameleo:
#175 
This issue will be broken into several parts for example:
- One issue to write missing docstrings
- One for removing unused imports and variables
- So on

Razzeee:
Sounds sensible.

My preferred setup would be to rely on the standard configuration and not configuring it at all. But getting there will take time. Am I correct that we should be able to just delete the config file then and it will just use pylint defaults?

requirements.txt
```diff
@@ -1,3 +1,4 @@
+pylint
```
Rechi: This file lists only the runtime dependencies of the addon-checker. `pylint` isn't a runtime dependency.

mzfr:
@Razzeee Running pylint with default config is possible and actually isn't a big task to achieve because there aren't many errors left, the most amounts of errors are about `snake_case vs the camelCase` so we need to decide that and then make sure everyone use that. Other than that just the docstrings and `Do not use `len(SEQUENCE)` to determine if a sequence is empty` errors are there. Which I think won't be a problem to fix.

That being said I still think that running pylint on complete default isn't a good idea(at least in my opinion) because pylint gives quite a few suggestions of itself and sometime can have `False-positives`.

LordGameleo:
@Razzeee it is possible and I am also suggesting that. As we progress with individual issues we will remove them one by one from disabled list. And once all these issues are solved we will be left with the default configuration. Its like we will be doing it in progressive way.

requirements.txt
LordGameleo: Then I will add this in travis script.

.travis.yml
```diff
@@ -91,8 +92,8 @@ script:
       python3 -m kodi_addon_checker --branch=$BRANCH ../repo;
     fi
   - if [[ -z "$REPO" || -z "$BRANCH" ]]; then
-      py.test -v;
+      pylint * --rcfile=.pylintrc;
     fi
   - if [[ -z "$REPO" || -z "$BRANCH" ]]; then
-      pycodestyle . --max-line-length=120;
+      py.test -v;
```
Rechi: What's the reason for running the code checker before the tests?

LordGameleo: kodi checker will create log files

LordGameleo:
> I disabled these checks to get a 10 on 10 score for now.
> 
> ```
>         broad-except,
>         too-few-public-methods,
>         parse-error,
>         invalid-name,
>         attribute-defined-outside-init,
>         pointless-statement,
>         useless-object-inheritance,
>         no-else-return,
>         no-name-in-module,
>         len-as-condition,
>         missing-docstring,
>         bad-whitespace,
>         duplicate-code,
>         unused-import,
>         unused-variable,
>         useless-else-on-loop,
>         bad-continuation,
>         import-error,
>         too-many-arguments,
>         logging-not-lazy,
>         undefined-variable,
>         syntax-error,
>         unidiomatic-typecheck,
>         too-many-nested-blocks,
>         inconsistent-return-statements,
>         arguments-differ,
>         consider-using-in,
>         deprecated-method,
>         redefined-outer-name,
> ```

@Rechi  I just disabled these checks which was causing the build to fail.... I used the feature of pylint which creates pylintrc file.  Everything except what I mentioned above is default. I used standard method of pylint to create it. We can remove all other things if we want to.

Rechi:
This is now a more reasonable list, then you have in the current `.pylintrc` file.
The reason why parse-error, pointless-statement, syntax-error and undefined-variable have to be ignored are not actual code errors, but linting LICENSE, MANIFEST.in, PUBLISH.md, requirements.txt and script.test.
Also I don't have to exclude no-name-in-module, bad-whitespace and import-error.

LordGameleo:
@Rechi for confirming.... Do you want me to make my own customized version of pylintrc file which includes only these errors which causes the build to fail. 
And yes, we are ignoring these errors just so we can merge this branch successfully without failing travis build. We will solve these errors in new issues and when they are those issues are solved we will remove those disabled errors from pylintrc( specific to these disabled ones ) and we will keep ignoring those errors which we don't have to ignore.

.travis.yml
LordGameleo: @Rechi  Should I run py.test before?

requirements.txt
Rechi: It's back again.

LordGameleo:
@Razzeee @mzfr @Rechi Is there anything to change here?

.pylintrc
```diff
@@ -0,0 +1,37 @@
+[MESSAGES CONTROL]
+
+disable=broad-except,
+        too-few-public-methods,
+        parse-error,
```
Rechi: This is only needed, because `script.test` is linted.
```
[Errno 2] No such file or directory: 'script.test/__init__.py' (parse-error)
```

.pylintrc
```diff
@@ -0,0 +1,37 @@
...
+        invalid-name,
+        attribute-defined-outside-init,
+        pointless-statement,
```
Rechi: This is only needed, because `requirements.txt` is linted.
```
requirements.txt:1:0: W0104: Statement seems to have no effect (pointless-statement)
requirements.txt:2:0: W0104: Statement seems to have no effect (pointless-statement)
requirements.txt:3:0: W0104: Statement seems to have no effect (pointless-statement)
requirements.txt:4:0: W0104: Statement seems to have no effect (pointless-statement)
requirements.txt:5:0: W0104: Statement seems to have no effect (pointless-statement)
requirements.txt:6:0: W0104: Statement seems to have no effect (pointless-statement)
```

.pylintrc
```diff
@@ -0,0 +1,37 @@
...
+        bad-whitespace,
+        duplicate-code,
+        unused-import,
+        unused-variable,
+        useless-else-on-loop,
+        bad-continuation,
+        import-error,
+        too-many-arguments,
+        logging-not-lazy,
+        undefined-variable,
```
Rechi: This is only needed, because `requirements.txt` is linted.
```
requirements.txt:1:0: E0602: Undefined variable 'pillow' (undefined-variable)
requirements.txt:2:0: E0602: Undefined variable 'polib' (undefined-variable)
requirements.txt:3:0: E0602: Undefined variable 'requests' (undefined-variable)
requirements.txt:4:0: E0602: Undefined variable 'radon' (undefined-variable)
requirements.txt:5:0: E0602: Undefined variable 'urllib3' (undefined-variable)
requirements.txt:6:0: E0602: Undefined variable 'xmlschema' (undefined-variable)
```

.pylintrc
```diff
@@ -0,0 +1,37 @@
...
+        syntax-error,
```
Rechi: This is only needed, because `LICENSE`, `MANIFEST.in`, `PUBLISH.md` and `README.md` are linted.
```
LICENSE:1:0: E0001: unexpected indent (<unknown>, line 1) (syntax-error)
MANIFEST.in:1:0: E0001: invalid syntax (<unknown>, line 1) (syntax-error)
PUBLISH.md:2:0: E0001: invalid syntax (<unknown>, line 2) (syntax-error)
README.md:1:0: E0001: invalid syntax (<unknown>, line 1) (syntax-error)
```

LordGameleo:
> This is now a more reasonable list, then you have in the current `.pylintrc` file.
> The reason why parse-error, pointless-statement, syntax-error and undefined-variable have to be ignored are not actual code errors, but linting LICENSE, MANIFEST.in, PUBLISH.md, requirements.txt and script.test.
> Also I don't have to exclude no-name-in-module, bad-whitespace and import-error.

@Rechi I am sorry I guess I couldn't get what you exactly said here. I got it now, I will fix these. Sorry for making you do the extra work

LordGameleo:
> This is now a more reasonable list, then you have in the current `.pylintrc` file.
> The reason why parse-error, pointless-statement, syntax-error and undefined-variable have to be ignored are not actual code errors, but linting LICENSE, MANIFEST.in, PUBLISH.md, requirements.txt and script.test.
> Also I don't have to exclude no-name-in-module, bad-whitespace and import-error.

@Rechi I am sorry I guess I couldn't get what you exactly said here. I got it now, I will fix these. Sorry for making you do the extra work

LordGameleo:
@Rechi  I came up with a better alternative. We can just check .py files rather than going for all.

.pylintrc
LordGameleo: How can we correct this?

Rechi: Do not lint `script.test`. It is only random Python code, which is used for addon-checker tests.

LordGameleo: Should we keep this disabled?

LordGameleo: or should we create __init__.py file?

Rechi: > Do not lint script.test. It is only random Python code, which is used for addon-checker tests.

Neither disable `parse-error` check or create a `__init__.py` file.

LordGameleo: Okay I did that.

LordGameleo:
@Rechi Check this once and tell me if you see anything now.

Rechi:
`pip install pylint` is missing in `.travis.yml`.

.pylintrc
```diff
@@ -0,0 +1,44 @@
...
+
+[MESSAGES CONTROL]
+
+disable=broad-except,
+        too-few-public-methods,
+        invalid-name,
+        attribute-defined-outside-init,
+        useless-object-inheritance,
+        no-else-return,
+        no-name-in-module,
```
Rechi: I do not have to disable this check.

.pylintrc
```diff
@@ -0,0 +1,44 @@
...
+        len-as-condition,
+        missing-docstring,
+        bad-whitespace,
```
Rechi: same

.pylintrc
```diff
@@ -0,0 +1,44 @@
...
+        duplicate-code,
+        unused-import,
+        unused-variable,
+        useless-else-on-loop,
+        bad-continuation,
+        import-error,
```
Rechi: same

.travis.yml
```diff
@@ -94,5 +94,5 @@ script:
       py.test -v;
     fi
   - if [[ -z "$REPO" || -z "$BRANCH" ]]; then
-      pycodestyle . --max-line-length=120;
+      pylint * --rcfile=.pylintrc;
```
Rechi: `--rcfile=.pylintrc` is the default, so it is not essential.

.pylintrc
LordGameleo: It is a problem with pylint when using virtual env. That's why it was showing this error in my pylint check. I am removing this for now from disabled list.
https://github.com/PyCQA/pylint/issues/1524

Rechi: Although I don't like to disable `no-name-in-module` and `import-error`, I see the reason now in travis log.
Please readd it.

.pylintrc
```diff
@@ -0,0 +1,43 @@
...
+       MANIFEST.in,
+       PUBLISH.md,
+       README.md,
+       requirements.txt,
+       script.test,
+
+
+[MESSAGES CONTROL]
+
+disable=broad-except,
```
Razzeee: Can we break this up in two categories just visually.

```
disable=stuff-that-we-actually-want-ignored,
       stuff-that-we-actually-want-ignored2,

       stuff-thats-only-ignored-for-now-and-will-be-addressed-in-a-pr,
       stuff-thats-only-ignored-for-now-and-will-be-addressed-in-a-pr2

```

LordGameleo: @Razzeee I think mostly all f them are ones which we are ignoring temporarily except import-error and no-name-in-module as It will cause build to fail due to virtual environment (as mentioned in https://github.com/PyCQA/pylint/issues/1524 )

LordGameleo:
@Razzeee @Rechi Please Review :)

.pylintrc
```diff
@@ -0,0 +1,51 @@
...
+
+# Types of check we want to disable
+# format
+# disable=stuff-that-we-actually-want-ignored,
+#        stuff-that-we-actually-want-ignored2,
+#
+#        stuff-thats-only-ignored-for-now-and-will-be-addressed-in-a-pr,
+#        stuff-thats-only-ignored-for-now-and-will-be-addressed-in-a-pr2
+disable=import-error,
+        no-name-in-module,
```
Rechi: I wouldn't classify those checks as wanted to be ignored.

LordGameleo: What all checks should be ignored then?

LordGameleo: @Rechi

PR #189, opened 2019-03-20
allow Python 3 compatibility check regardless of branch

anxdpanic:
YouTube, and Twitch are both Python 3 compatible but due to residing in the Isengard branch the Python 3 compatibility checks are not applied and running the checks against a higher branch results in xml schema issues.

This would allow me to apply the Python 3 checks on PRs to the add-on's repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,4 +96,5 @@ You can use the tool with the following options:
 --PR                        only when the tool is running on a pull request
 --allow-folder-id-mismatch  allow the addon's folder name and id to mismatch
 --reporter                  enable a reporter, this option can be used multiple times
+--check-python3             include Python 3 compatibility check regardless of the current branch
 ```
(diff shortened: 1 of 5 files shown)


anxdpanic:
Realized I forgot to add to readme, was the only change made in the last push. Sorry about that.

Rechi:
I do not like this special command line argument.

The reason why some checks are disabled for older Kodi versions is that they are not compatible with Python 2.6, not required in all cases or not fully correct.

Which of the optional Python checks do you want?

anxdpanic:
I was wanting to run the python3 compatibility checks against the add-ons mentioned, and figured others may find it useful. We can close this if that's unwanted.

anxdpanic:
I see why this could cause issues, this is probably better done with a separate test. 👍

PR #190, opened 2019-03-21
repository: only un-gzip if url ends with .gz

Rechi:


Rechi:
Line 24 is the if statement.
Are you sure there should be an empty line between the if statement and the code of if the block?

LordGameleo:
No I mean after if statement

Issue #198, opened 2019-07-17
Feature: Ignore .tox and test/ directories


dagwieers:
Currently the tool takes all files into account, even files that are not part of the plugin itself (like supporting infrastructure, its own kodi-addon-checker logfiles or build tools).

The add-on should not assume all files are part of the plugin, and should stick to the files that actually belong to what is distributed. Either by listing those files only (starting from addon.xml) or by accepting a ZIP file instead (so we can first create the ZIP file and then provide it to the tool).

In any case, it should exclude .tox/ and test/ directories to start with so people can use it as part of their development cycle.

dagwieers:
Example output that should not be a problem:
```
ERROR: ./plugin.video.vrt.nu/Build-Zip.ps1 is marked as stand-alone executable
ERROR: ./plugin.video.vrt.nu/test/run.py is marked as stand-alone executable
--snip--
WARN: Found non whitelisted file ending in filename ./plugin.video.vrt.nu/kodi-addon-checker-report.log
WARN: Found non whitelisted file ending in filename ./plugin.video.vrt.nu/kodi-addon-checker.log
WARN: Found non whitelisted file ending in filename ./plugin.video.vrt.nu/Build-Zip.ps1
WARN: Found non whitelisted file ending in filename ./plugin.video.vrt.nu/test/userdata/vrtloginrt.tkn
WARN: Found non whitelisted file ending in filename ./plugin.video.vrt.nu/test/userdata/XVRTToken.tkn
WARN: Found non whitelisted file ending in filename ./plugin.video.vrt.nu/test/userdata/ondemand_vrtPlayerToken.tkn
WARN: Found non whitelisted file ending in filename ./plugin.video.vrt.nu/test/userdata/live_vrtPlayerToken.tkn
WARN: Found non whitelisted file ending in filename ./plugin.video.vrt.nu/test/userdata/user_XVRTToken.tkn
```

anxdpanic:
For the logs, run from out of directory.
With travis, most could be resolved as well. ex. https://github.com/jdf76/plugin.video.youtube/blob/master/.travis.yml 
https://travis-ci.org/jdf76/plugin.video.youtube/builds/558284685
Cloning master may not be wanted but is easily changed.

Razzeee:
[...]

PR #200, opened 2019-07-31
only create kodi-addon-checker-report.log when using --reporter=log

anxdpanic:
resolves one of the log files mentioned in #198


Razzeee:
Thank you

PR #208, opened 2020-03-08
Prepare xbmc.python for matrix

enen92:
This will make the addon-checker report a problem if submitting addons to matrix with versions lower than 3.0.0

@Razzeee it would be nice if you could create a new release.


Razzeee:
published

enen92:
Thanks @Razzeee

Issue #207, opened 2019-11-15
Add option to prevent checking if addon already exists in Kodi addon repository


goggle:
This tools tries to establish a internet connection to check if the addon is already available in Kodi's addon repository.
If establish that connection fails, I get this:
```
Traceback (most recent call last):
  File "/tmp/kodi-check/lib/python3.7/site-packages/urllib3/response.py", line 425, in _error_catcher
    yield
  File "/tmp/kodi-check/lib/python3.7/site-packages/urllib3/response.py", line 507, in read
    data = self._fp.read(amt) if not fp_closed else b""
  File "/usr/lib64/python3.7/http/client.py", line 457, in read
    n = self.readinto(b)
  File "/usr/lib64/python3.7/http/client.py", line 501, in readinto
    n = self.fp.readinto(b)
  File "/usr/lib64/python3.7/socket.py", line 589, in readinto
    return self._sock.recv_into(b)
  File "/usr/lib64/python3.7/ssl.py", line 1071, in recv_into
    return self.read(nbytes, buffer)
  File "/usr/lib64/python3.7/ssl.py", line 929, in read
    return self._sslobj.read(len, buffer)
socket.timeout: The read operation timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/tmp/kodi-check/lib/python3.7/site-packages/requests/models.py", line 750, in generate
    for chunk in self.raw.stream(chunk_size, decode_content=True):
  File "/tmp/kodi-check/lib/python3.7/site-packages/urllib3/response.py", line 564, in stream
    data = self.read(amt=amt, decode_content=decode_content)
  File "/tmp/kodi-check/lib/python3.7/site-packages/urllib3/response.py", line 529, in read
    raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
  File "/usr/lib64/python3.7/contextlib.py", line 130, in __exit__
    self.gen.throw(type, value, traceback)
  File "/tmp/kodi-check/lib/python3.7/site-packages/urllib3/response.py", line 430, in _error_catcher
    raise ReadTimeoutError(self._pool, None, "Read timed out.")
urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='kodi.mirror.wearetriple.com', port=443): Read timed out.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/tmp/kodi-check/bin/kodi-addon-checker", line 8, in <module>
    sys.exit(main())
[...]

PR #211, opened 2020-03-08
increase repository request timeout to 30 seconds

anxdpanic:
This should reduce the connection timeout issues on PR's
The connection to the failing mirrors have required ~20 seconds in the occurrences I've checked.


anxdpanic:
https://travis-ci.org/xbmc/addon-check/jobs/659878906 is an example of the issue this should address.

enen92:
Partially addresses https://github.com/xbmc/addon-check/issues/207 too

PR #212, opened 2020-03-08
Add checks for all missing asset art types

enen92:
As a result of https://github.com/xbmc/repo-scripts/pull/1299 we need to start checking the other art types missing because they can break our repository generator script.

We discussed and agreed on the new rules for clearlogo, banner and screenshot asset types in https://forum.kodi.tv/showthread.php?tid=351853. Wiki has already been updated too:

https://kodi.wiki/view/Addon.xml
https://kodi.wiki/view/Add-on_structure#resources.2Ffanart.jpg

### DIFF
--- a/kodi_addon_checker/check_artwork.py
+++ b/kodi_addon_checker/check_artwork.py
@@ -88,7 +88,7 @@ def check_artwork(report: Report, addon_path: str, parsed_xml, file_index: list,
 def _check_image_type(report: Report, asset: tuple, parsed_xml, addon_path: str, kodi_version: KodiVersion):
     """Check for whether the given image type exists or not if they do """
 
-    fallback, images = _assests(asset.image_type, parsed_xml, addon_path)
+    fallback, images = _assets(asset.image_type, parsed_xml, addon_path)
 
     for image in images:
         if image:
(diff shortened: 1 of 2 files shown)


kodi_addon_checker/check_artwork.py
```diff
@@ -26,9 +27,53 @@ def check_artwork(report: Report, addon_path: str, parsed_xml, file_index: list,
...
+            {
+                'extension': ['.png'],
+                'transparency': False,
+                'sizes': [(256, 256), (512, 512)]
+            }
+        ),
+        Asset(
+            'fanart',
+            {
+                'extension': [".jpg", ".jpeg"],
```
anxdpanic: File size seems to be the reason to limit this, would it be possible to include png and set a max file size(150-200kb?)?
I'm using a 3840x2160 .png for the Twitch add-on which is only 32kb
https://github.com/MrSprigster/Twitch-on-Kodi/blob/master/fanart.png

enen92: Probably not a bad idea to not limit so much on formats and use the root cause of the specification: filesize. I'll see what I can do.

enen92: Thinking about this a bit more the filesize check should be an easy one. Allowing fanarts in png should only be done for krypton and above versions, gotham-krypton have the fanart.jpg and icon.png fallback in the root of the addon, so we need to somehow relax the check while not overcomplicating the code (the linter is pretty anoying on this project :P)

kodi_addon_checker/check_artwork.py
```diff
@@ -26,9 +27,53 @@ def check_artwork(report: Report, addon_path: str, parsed_xml, file_index: list,
...
+                'transparency': None,
```
Razzeee: Should this really be `None` all others are using `False`

Razzeee: oh, fanart are okay to have transparency

PR #215, opened 2020-03-25
[check_artwork] add file size check to artwork

anxdpanic:


anxdpanic:
I went with 750kb for fanart and screenshots to start, would like feedback on what these should be set to

Razzeee:
While I think the default is fine, it seems to mostly fail on the skins repo, might make sense to ask some skinners on their opinions.

kodi_addon_checker/check_artwork.py
```diff
@@ -35,39 +35,44 @@ def check_artwork(report: Report, addon_path: str, parsed_xml, file_index: list,
...
-                'sizes': [(256, 256), (512, 512)]
+                'sizes': [(256, 256), (512, 512)],
+                'max_file_size': None
             }
         ),
         Asset(
             'fanart',
             {
-                'extension': [".jpg", ".jpeg"],
+                'extension': [".jpg", ".jpeg", ".png"],
```
enen92: this has to be done differently. The checker is available to all kodi versions since gotham and only since krypton the assets tag was implemented. Before kodi has fanart.jpg hardcoded in the root of the addon. So this has to depend on the Kodi version => [".jpg", ".jpeg", ".png"] only for krypton++; [".jpg", ".jpeg"] for gotham-krypton

anxdpanic:
I've asked for feedback from the skinners. I will update the PR with the fix to the krypton issue as well after I've heard back.

kodi_addon_checker/check_artwork.py
anxdpanic: png should be krypton+ now

anxdpanic:
@sualfred was the only feedback I've received so far. 
750kb to 1mb is hopefully a good to place to start. 
I set screenshots(1080 max) @ 750kb, and fanart(4k max) @ 1000kb

kodi_addon_checker/check_artwork.py
```diff
@@ -29,45 +29,55 @@ def check_artwork(report: Report, addon_path: str, parsed_xml, file_index: list,
...
-                'sizes': [(256, 256), (512, 512)]
+                'sizes': [(256, 256), (512, 512)],
+                'max_file_size': None
             }
         ),
         Asset(
             'fanart',
             {
-                'extension': [".jpg", ".jpeg"],
+                'extension': fanart_extensions,
```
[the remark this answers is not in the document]
anxdpanic: I've moved the condition inline

anxdpanic:
It looks like there are a lot of `too large` in the repository tests. 
Should this be set to `WARNING` instead of `PROBLEM` with a `TODO` or hold off, until we have a better picture of what these sizes should be in practice?

I can always convert my fanart to jpg for now.

enen92:
Matrix is the important branch now, as long as it doesn't error out there I'm fine in enforcing it in lower branches

kodi_addon_checker/check_artwork.py
mediaminister: If the `<assets>` tag works on Krypton, why is png not allowed for Krypton?

PR #216, opened 2020-03-26
Improve addon-checker version comparison

enen92:
To fix https://github.com/xbmc/repo-scripts/pull/1356

`distutils.version` is not a robust way to do version comparison. This has been already raised to python core and the patch was denied: https://bugs.python.org/issue30272

This changes the implementation to rely on `packaging.version` which has a much more robust algorithm.

This PR also changes `KodiVersion.py` to `versions.py` for consistency. `LooseVersion` is now `AddonVersion` and centralized on `versions.py`. Tests added to make sure this won't break in the future.

### DIFF
--- a/kodi_addon_checker/addons/AddonDependency.py
+++ b/kodi_addon_checker/addons/AddonDependency.py
@@ -7,8 +7,7 @@
 """
 
 import xml.etree.ElementTree as ET
-from distutils.version import LooseVersion
-
+from ..versions import AddonVersion
 
 class AddonDependency():
     def __init__(self, import_xml: ET.Element):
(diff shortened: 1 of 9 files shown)


anxdpanic:
Looks like this broke one of the [tests](https://github.com/xbmc/addon-check/pull/215/files#diff-59e284dc6e956a7e8a094f8e595dc827R47), and the pylint is minor

Should `packaging` be added to requirements.txt? Don't have time to check right atm

enen92:
Packaging should be part of setuptools so I doubt it will be ever needed as an external package. I'll fix the tests

enen92:
Travis is green now. No errors from this change in all branches.
@Razzeee it'd be nice to have a new release

anxdpanic:
@enen92 Looks like a minor issue with reporting required version of a dependency,
https://travis-ci.org/github/MrSprigster/Twitch-on-Kodi/builds/668410076#L269

PR #217, opened 2020-03-26
Retry unreliable repositories

dagwieers:
If a repository times out or has other connectivity issues, this change ensures the request is retried, trying other repositories when available (round robin).

### DIFF
--- a/kodi_addon_checker/addons/Repository.py
+++ b/kodi_addon_checker/addons/Repository.py
@@ -20,7 +20,14 @@ class Repository():
         super(Repository, self).__init__()
         self.version = version
         self.path = path
-        content = requests.get(path, timeout=(30, 30)).content
+
+        # Recover from unreliable mirrors
+        session = requests.Session()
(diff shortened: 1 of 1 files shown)


dagwieers:
We often have problems in our GitHub CI actions because of the below:

```
requests.exceptions.ConnectionError: HTTPConnectionPool(host='mirror.ufs.ac.za', port=80): Read timed out.
```

And we see this happen often for the same set of mirrors, making our tests unreliable.
https://github.com/add-ons/plugin.video.vrt.nu/pull/713/checks

enen92:
Thanks

dagwieers:
So apparently **download.kodi.tv** has its certificate expired, and it retries 5 times the same host. A DNS round-robin setup could be useful to also recover from these kind of issues.

enen92:
@dagwieers @yol fixed the issue

yol:
@dagwieers Thanks for the pointer

yol:
> A DNS round-robin setup could be useful to also recover from these kind of issues.

That would not be compatible with mirrorbits, right?

PR #222, opened 2020-04-07
Added a check to prevent changelog.txt files in Leia and up

basrieter:
We motivate devs to not use the changelog.txt anymore but the <news> tag in the addon.xml

### DIFF
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -89,7 +89,7 @@ def start(addon_path, args, all_repo_addons, config=None):
                 check_files.check_for_new_language_directory_structure(addon_report, addon_path, supported=False)
 
             if KodiVersion(args.branch) >= KodiVersion("krypton"):
-                check_files.check_for_legacy_changelog(addon_report, addon_path, args.branch)
+                check_files.check_for_legacy_changelog(addon_report, addon_path, parsed_xml, args.branch)
 
             check_string.check_for_invalid_strings_po(addon_report, file_index)
 
--- a/kodi_addon_checker/check_addon.py
+++ b/kodi_addon_checker/check_addon.py
@@ -88,6 +88,9 @@ def start(addon_path, args, all_repo_addons, config=None):
             else:
                 check_files.check_for_new_language_directory_structure(addon_report, addon_path, supported=False)
 
+            if KodiVersion(args.branch) >= KodiVersion("krypton"):
+                check_files.check_for_legacy_changelog(addon_report, addon_path, args.branch)
+
             check_string.check_for_invalid_strings_po(addon_report, file_index)
 
             # Kodi 18 Leia + deprecations
--- a/kodi_addon_checker/check_files.py
+++ b/kodi_addon_checker/check_files.py
@@ -91,17 +91,23 @@ def addon_xml_matches_folder(report: Report, addon_path: str, parsed_xml, folder
             report.add(Record(PROBLEM, "Addon id and folder name does not match."))
 
 
-def check_for_legacy_changelog(report: Report, addon_path: str, branch: str):
+def check_for_legacy_changelog(report: Report, addon_path: str, parsed_xml: ET.Element, branch: str):
     """Check whehter a legacy changelog.txt is present in the addon folder root"""
     changelog_path = os.path.join(addon_path, "changelog.txt")
+
     if os.path.isfile(changelog_path):
-        report.add(
-            Record(
-                PROBLEM if KodiVersion(branch) >= KodiVersion("leia") else WARNING,
-                "A changelog.txt file was found. Use the <news> tag in addon.xml for the most recent changes "
-                "instead of a changelog.txt. See https://kodi.wiki/view/Addon.xml"
+        news = parsed_xml.find("./extension/news")
+        news_present = news is not None and news.text
+
+        if not news_present:
+            report.add(
+                Record(
+                    PROBLEM if KodiVersion(branch) >= KodiVersion("krypton") else WARNING,
+                    "A changelog.txt file was found but no <news> tag was present in the addon.xml. Please add a "
+                    "<news> element to your addon.xml and include the most recent changes there. "
+                    "See https://kodi.wiki/view/Addon.xml"
+                )
             )
-        )
 
 
 def check_for_new_language_directory_structure(report: Report, addon_path: str, supported=True):
--- a/kodi_addon_checker/check_files.py
+++ b/kodi_addon_checker/check_files.py
@@ -16,6 +16,7 @@ from .common import relative_path
 from .common.decorators import posix_only
 from .record import INFORMATION, PROBLEM, WARNING, Record
 from .report import Report
+from .versions import KodiVersion
 
 
 def check_for_invalid_xml_files(report: Report, file_index: list):
@@ -88,6 +89,19 @@ def addon_xml_matches_folder(report: Report, addon_path: str, parsed_xml, folder
                                            "to Kodi's official repository.".format(addon_id)))
         else:
             report.add(Record(PROBLEM, "Addon id and folder name does not match."))
+
+
+def check_for_legacy_changelog(report: Report, addon_path: str, branch: str):
+    """Check whehter a legacy changelog.txt is present in the addon folder root"""
+    changelog_path = os.path.join(addon_path, "changelog.txt")
+    if os.path.isfile(changelog_path):
+        report.add(
+            Record(
+                PROBLEM if KodiVersion(branch) >= KodiVersion("leia") else WARNING,
+                "A changelog.txt file was found. Use the <news> tag in addon.xml for the most recent changes "
+                "instead of a changelog.txt. See https://kodi.wiki/view/Addon.xml"
+            )
+        )
 
 
 def check_for_new_language_directory_structure(report: Report, addon_path: str, supported=True):


kodi_addon_checker/check_addon.py
```diff
@@ -86,6 +86,9 @@ def start(addon_path, args, all_repo_addons, config=None):
             if KodiVersion(args.branch) >= KodiVersion("isengard"):
                 check_files.check_for_legacy_language_path(addon_report, addon_path)
 
+            if KodiVersion(args.branch) >= KodiVersion("leia"):
```
enen92, line 89: Should be krypton: https://kodi.wiki/view/Add-on_structure#changelog.txt

kodi_addon_checker/check_files.py
```diff
@@ -89,6 +89,14 @@ def addon_xml_matches_folder(report: Report, addon_path: str, parsed_xml, folder
             report.add(Record(PROBLEM, "Addon id and folder name does not match."))
 
 
+def check_for_legacy_changelog(report: Report, addon_path: str):
+    """Check whehter a legacy changelog.txt is present in the addon folder root"""
+    changelog_path = os.path.join(addon_path, "changelog.txt")
+    if os.path.isfile(changelog_path):
+        report.add(Record(PROBLEM, "A changelog.txt file was found. Use the <news> tag in addon.xml "
```
enen92, line 96: I am not sure others will agree with the problem. @razzeee, @rechi what do you guys think?

kodi_addon_checker/check_addon.py
basrieter: Indeed, I wasn't sure. But I will change that.

kodi_addon_checker/check_files.py
basrieter: We can make it fail for Leia and Matrix, and warn for Krypton?

enen92: That would be better. there are still a lot of addons in krypton using the old changelog, I'm not confortable with hardfailing the addon-checker because of that. Matrix and Leia don't have much addons and we've been enforcing it anyway.

Razzeee: I'm actually using both. As news only should include the latest changes. But the changelog will even have older versions.

enen92: @Razzeee the thing is the changelog file tends to grow when you create new releases and is a helper file that is not parsed by kodi anymore, so there's no reason to be shipping it from the mirrors and making addon packages bigger. Ideally you should be able to use anyfiles you want in your upstream repo but avoid submitting it to the repo (adding to .gitattributes for example)

Razzeee: I think this is getting more and more complicated. Maybe we need something like `npm`, that packages your release. Maybe that's even something this project could do.

Not sure how .gitattributes would help with that.

enen92: @Razzeee see for instance https://github.com/enen92/plugin.video.rtpplay/commit/1a2bd4c8825b0cb11b9edb0e666dd6cb567f0f05

The addon-submitter by romanv uses `git archive` to create the addon package before submitting automatically to the repo. If you blacklist the files you don't want to have packaged in .gitignore they won't be sent on the pull request. But yeah I agree, this somehow is getting complicated but at the same time we're working on creating a simple workflow to get everything done with little effort

enen92: A possible option could be to merge the two projects. Also include them when using the kodi-addon-generator

basrieter: Indeed @Razzeee: if you configure the .gitattributes correctly (https://github.com/retrospect-addon/plugin.video.retrospect/blob/master/.gitattributes) then the `git archive` and the GitHub `Download zip` basically generate your release. It really saves a lot of time.

Razzeee: I'll need to read up on `git archive` thank you

basrieter: I updated the code to only warn for Krypton and lower, and error for Leia and up.

kodi_addon_checker/check_files.py
```diff
@@ -89,6 +90,18 @@ def addon_xml_matches_folder(report: Report, addon_path: str, parsed_xml, folder
             report.add(Record(PROBLEM, "Addon id and folder name does not match."))
 
 
+def check_for_legacy_changelog(report: Report, addon_path: str, branch: str):
+    """Check whehter a legacy changelog.txt is present in the addon folder root"""
+    changelog_path = os.path.join(addon_path, "changelog.txt")
+    if os.path.isfile(changelog_path):
+        if KodiVersion(branch) < KodiVersion("leia"):
```
enen92, line 97: please use a single line:

```python
report.add(
    Record(
        ERROR if KodiVersion(branch) >= KodiVersion("leia") else WARNING
       "A changelog.txt file was found. Use the <news> tag in addon.xml "
                                       "for the most recent changes instead of a changelog.txt."))
```

enen92: Maybe point to the wiki link for `<news>` in the message so that users know what to do

basrieter: Done

kodi_addon_checker/check_files.py
```diff
@@ -89,6 +90,19 @@ def addon_xml_matches_folder(report: Report, addon_path: str, parsed_xml, folder
             report.add(Record(PROBLEM, "Addon id and folder name does not match."))
 
 
+def check_for_legacy_changelog(report: Report, addon_path: str, branch: str):
+    """Check whehter a legacy changelog.txt is present in the addon folder root"""
```
Razzeee, line 94:
```suggestion
    """Check whether a legacy changelog.txt is present in the addon folder root"""
```

kodi_addon_checker/check_files.py
```diff
@@ -90,6 +91,19 @@ def addon_xml_matches_folder(report: Report, addon_path: str, parsed_xml, folder
...
+    changelog_path = os.path.join(addon_path, "changelog.txt")
+    if os.path.isfile(changelog_path):
+        report.add(
+            Record(
+                PROBLEM if KodiVersion(branch) >= KodiVersion("leia") else WARNING,
```
enen92, line 100: Probably not a good idea to report it as problem. Maybe turn this into a warning only f the news is empty? cc @ronie

ronie: well, if you can point me to 1 online article that says it is not a good idea to accompany your software with a changelog file, then i'm all ears ;-)

enen92:
So @basrieter I think it's probably a good ideia to re-write this and only `WARN` if a changelog.txt is found but the `<news>` element is empty on addon.xml. What do you think?

basrieter:
So if a changelog is present but no news tag is present?

I would say that should be an error. If there is a changelog then put the basics in the news tag. That is the only way to show it in the GUI in Leia and Matrix.

 If there is no changelog, the news tag can be left empty.

basrieter:
So I updated it with this behaviour:

- If a changelog.txt is present but no <news> tag, it will cause an error on Leia and Matrix
- If a changelog.txt is present and a <news> tag, then it will be a warning with a suggestion to remove the changelogt.txt

basrieter:
@ronie is this new option an acceptable one?

ronie:
the point i was trying to make is it's common practice (and recommended) for software to include a changelog file.
we should not be issuing a warning for that imo.

basrieter:
> the point i was trying to make is it's common practice (and recommended) for software to include a changelog file.
> we should not be issuing a warning for that imo.

Ok, understandable. I will remove the warning then.

enen92:
@sualfred the anti-news guy. Can you give some feedback?

sualfred:
Anti-news guy here: 
Was there a really good reason for dropping the changelog.txt behaviour in Kodi? Never understood it and it was before I was in the team.

I hate the news tag because it's
a) I don't see a reason why a changelog should be in a addon definition file. Maybe optional for reading a "latest change" in our addon browser or something. Anything else will be look awful. Especially if you want to keep track and provide a long and complete version history.
b) Kodi caused issues with indentation of the .xml files. At least it did it in Krypton and there were tons of white spaces in the textviewer dialog. Maybe because some uses tabs or spaces? Dunno. Never touched it again.
c) Everyone is used to have a changelog.txt. It's common used everywhere since decades in all kinds of software or update packages. Personally I always check for those files to see what has been changed in the past.

But this are just my 2cts. I wish me the old changelog.txt reading back in Kodi. Especially because people never really switched over to the addon.xml method. Only a few are using it. People are used to have a changelog.txt while browsing GitHub and other sources.  

Btw: I've implemented a helper function in my `script.embuary.helper` to have a changelog.txt content available in a skin. I'm a rebel ;)

enen92:
@sualfred

dear anti-news guy, with this change:

- changelog.txt can be submitted with any addon (no problem including it)
- developers can not use the news tag at all (as long as a changelog is not submitted too)
- if a changelog is present and the author hasn't filled the news tag...it gives an error.

Are you ok with this behaviour?

sualfred:
No, news tag + changelog.txt should be allowed and optional. No matter if one of it is missing or present.

enen92:
^^ @basrieter looks like we have 2 similar opinions here. Can you change the PROBLEM to WARNING?

basrieter:
The thing that I don't like is: that with a changelog present and no `<news>`, the information is not shown in Kodi at all. If there is a changlog, it's a small effort to copy the most recent changes into the `<news>` tag and submit it. That way users can see it. So that's why I think it should be an error. 

But this aside: I actually script the replacement of it's content with that of the changelog.txt during build and could also very much live without it as long as the changelog.txt/md is displayed again. So why don't we revert that change in Kodi? Or am I suggestion a ludicrous idea?

sualfred:
+1 from me. Reverting it.

ronie:
i'd like to join the anti-news movement, if that's ok with @sualfred 

the choice to drop changelog in favour of news certainly wasn't agreed upon unanimously,
and the reason for the change were pretty mood imo.
but, it was mostly the team skinners who liked it at the time.

https://github.com/xbmc/xbmc/pull/9351

sualfred:
Still see no real benefit. And I'm sitting on both chairs. Skinner and Python monkey.

Anyway, @HitcherUK 's comment is the only one which makes sense:
https://github.com/xbmc/xbmc/pull/9351#issuecomment-196897356

My suggestion is:
- keeping <news> for the most recent changes to have them available as `ListItem.AddonNews` and to show this information on our webpage addon browser maybe. This still gives skinners the option to show the information if they really want to.
- change the addon info dialog button back to open a existing changelog.txt

basrieter:
> Still see no real benefit. And I'm sitting on both chairs. Skinner and Python monkey.
> 
> Anyway, @HitcherUK 's comment is the only one which makes sense:
> [xbmc/xbmc#9351 (comment)](https://github.com/xbmc/xbmc/pull/9351#issuecomment-196897356)
> 
> My suggestion is:
> 
> * keeping `<news>`  for the most recent changes to have them available as `ListItem.AddonNews` and to show this information on our webpage addon browser maybe. This still gives skinners the option to show the information if they really want to.
> * change the addon info dialog button back to open a existing changelog.txt

Ok, let me look into that. Don't know if it will be a lot of work, but I will see. Anyone else that we need to get onboard for this?

Razzeee:
>a) I don't see a reason why a changelog should be in a addon definition file. Maybe optional for reading a "latest change" in our addon browser or something. Anything else will be look awful. Especially if you want to keep track and provide a long and complete version history.

@sualfred as far as I remember this was always about "latest change" nothing else. you make it seem like it wasn't. The addon-checker even enforces a max lenght on "news", if I remember correct. So the recommendation always was, to have your changelog and fill the news tag with the most interesting stuff, in the best case even explicitly tailored and targeted for end user consumption.

Going back reading that merge request, I kinda think it would have been nice to have this translatable. (but yeah translations are broken anyway)


I guess we should ping @tamland out of courtesy.

enen92:
Closing. The anti-news movement has won :)
Thanks for the contribution @basrieter

PR #223, opened 2020-04-07
Don't check for file permissions on Windows

basrieter:
There is no need to check for file permissions on windows. It now shows a lot of errors because all files are executable on Windows.

### DIFF
--- a/kodi_addon_checker/check_files.py
+++ b/kodi_addon_checker/check_files.py
@@ -13,6 +13,7 @@ import xml.etree.ElementTree as ET
 
 from . import handle_files
 from .common import relative_path
+from .common.decorators import posix_only
 from .record import INFORMATION, PROBLEM, WARNING, Record
 from .report import Report
 
@@ -129,6 +130,7 @@ def check_file_whitelist(report: Report, file_index: list, addon_path: str):
                                   relative_path(os.path.join(file["path"], file["name"]))))
 
 
+@posix_only
 def check_file_permission(report: Report, file_index: list):
     """Check whether the files present in addon are marked executable
        or not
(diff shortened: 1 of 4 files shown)


kodi_addon_checker/check_files.py
```diff
@@ -135,6 +135,9 @@ def check_file_permission(report: Report, file_index: list):
         :file_index: list having names and path of all the files present in addon
     """
 
+    if os.name == "nt":
```
enen92, line 138: Shouldn't the check be done outside of the method? Or decorate it with a windows only helper wrapper...
Seems silly to enter the method just to return early. In python it might be seen as a bad practice or non-pythonic.
Either way I don't use windows and I am not the best person to review this

basrieter: I had it outside, but changed it. If we included it in the calling method, it should be 'aware' of the functionality inside the `check_file_permission` method and from an Object Orientation perspective I would say that it should not. That is why I kept it inside: `check_file_permission` actually knows what is being validated and can decide it is not needed for Windows.

But hey, one way or the other works for me.

enen92: why don't you anotate the method with something like a `@posix_only` decorator that would check `os.name` or `sys.platform` and in case of windows return early otherwise enter the method?

basrieter: Interesting, I will have a look at it tonight.

basrieter: I added a decorator to it (after some discussion with you on Slack)

basrieter:
Moving the `common.py` broke the unittests. Fixing that.

Issue #224, opened 2020-04-08
[bug] Broken addon version comparator with beta


SylvainCecchetto:
Since some days (maybe the 0.0.17 release) the add-on version comparator is broken with beta syntax.

My local addon.xml is `0.2.19~beta01` but I get a `ERROR: plugin.video.catchuptvandmore addon already exist with a higher version: 0.2.18 in krypton branch`.

This bug was not here before (at least, it worked with kodi-addon-checker 0.0.16)

Razzeee:
https://github.com/xbmc/addon-check/pull/216

enen92:
Thanks for reporting @SylvainCecchetto 

Problem is that `0.2.19~beta01` is not PEP404 compliant, hence the `packaging.version.parse` function parses it as a `packaging.version.LegacyVersion` object instead of a `packaging.version.Version` object. `LegacyVersion` is always lower than `Version` no matter what.

I gave this a test with the Kodi internal versioning system and your version is in fact higher than `0.2.18`, so the addon checker should mimic the kodi behaviour.

I'll try to come up with a fix asap.

PR #225, opened 2020-04-08
[Versions] Support legacy non-compliant PEP440 development versions 

enen92:
Fixes https://github.com/xbmc/addon-check/issues/224

Kodi supports non-PEP440 compliant version for development purposes, eg:

`1.0.1~beta`
`1.0.1~beta01`
etc

As those are not PEP440 compliant, the `packaging.version` module parses the version to `LegacyVersion` objects instead of `Version` objects. When the comparator compares both objects, `LegacyVersion` is always lower.

This is solved by ensuring that when a version string like those above is supplied to addon-checker (thus containing `~alpha` or `~beta`) and the resulting object is a `LegacyVersion`, the `AddonVersion` class does an implicit conversion to the PEP440 format: `x.x.xalphay`.

Tests added to mimic the ones in Kodi.

### DIFF
--- a/kodi_addon_checker/versions.py
+++ b/kodi_addon_checker/versions.py
@@ -6,13 +6,22 @@
     See LICENSES/README.md for more information.
 """
 
-from packaging.version import parse
+from packaging.version import parse, Version, LegacyVersion
 from kodi_addon_checker import ValidKodiVersions
 
 
 class AddonVersion():
     def __init__(self, version):
         self.version = parse(str(version))
+        # non PEP440 compliant versions (legacy), for beta and alpha versions
+        # convert them into PEP440 format: 1.1.0~beta01 -> 1.1.0beta01
+        if self._islegacy_devversion:
+            self.version = Version(version.replace("~", ""))
+
+    @property
+    def _islegacy_devversion(self):
+        return isinstance(self.version, LegacyVersion) and \
+            ("~beta" or "~alpha" in self.version)
 
     def __lt__(self, other):
         if not isinstance(other, self.__class__):
--- a/kodi_addon_checker/versions.py
+++ b/kodi_addon_checker/versions.py
@@ -12,7 +12,7 @@ from kodi_addon_checker import ValidKodiVersions
 
 class AddonVersion():
     def __init__(self, version):
-        self.version = parse(str(version))
+        self.version = parse(str(version).lower())
         # non PEP440 compliant versions (legacy), for beta and alpha versions
         # convert them into PEP440 format: 1.1.0~beta01 -> 1.1.0beta01
         if self._islegacy_devversion:
--- a/tests/test_versions.py
+++ b/tests/test_versions.py
@@ -24,5 +24,17 @@ def test_addonversion_kodi_simplevsstring():
     assert AddonVersion("0.5.4.1") > AddonVersion("0.5.4+matrix.1")
 
 
+def test_kodi_non_compliant_pep440_builtin_versions():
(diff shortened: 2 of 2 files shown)


enen92:
@SylvainCecchetto this should fix your issue. Thanks for reporting.

SylvainCecchetto:
Tested on my side, it works fine! Thank you!

tests/test_versions.py
```diff
@@ -18,11 +18,24 @@ def test_addonversion_simple_major():
...
+def test_kodi_non_compliant_pep440_builtin_versions():
+    # mimics the tests in TestAddonVersion.cpp (non PEP440 compliant)
+    # try to use PEP440 compliant versions whenever possible
+    assert AddonVersion("1.0.0-1") > AddonVersion("1.0.0")
+    assert AddonVersion("1.0.0~beta2") > AddonVersion("1.0.0~beta1")
+    assert AddonVersion("1.0.1~beta01") > AddonVersion("1.0.0")
+    assert AddonVersion("1.0.1~beta") > AddonVersion("1.0.0")
+    assert AddonVersion("1.0.0~beta") > AddonVersion("1.0.0~alpha")
+    assert AddonVersion("1.0.0~beta2") > AddonVersion("1.0.0~alpha")
+    assert AddonVersion("1.0.0~alpha3") > AddonVersion("1.0.0~alpha2")
```
Razzeee, line 37: Would tests with `+matrix.2` also make sense? Do we need to combine like `1.0.0+matrix.1~beta`?

kodi_addon_checker/versions.py
```diff
@@ -6,13 +6,22 @@
...
+        self.version = parse(str(version).lower())
+        # non PEP440 compliant versions (legacy), for beta and alpha versions
+        # convert them into PEP440 format: 1.1.0~beta01 -> 1.1.0beta01
+        if self._islegacy_devversion:
+            self.version = Version(version.replace("~", ""))
+
+    @property
+    def _islegacy_devversion(self):
+        return isinstance(self.version, LegacyVersion) and \
+            ("~beta" or "~alpha" in self.version)
```
Razzeee, line 24: I was under the impression that you could write whatever after `~` and I probably did at some point? At least for test builds, but I guess that was never supported by core?

tests/test_versions.py
enen92: I don't think it makes much sense. `+matrix.1` is just a "wrapper" for the version available in matrix, so `1.0.0+matrix.1` is theoretically equivalent (at least for us) to version `1.0.0`. According to PEP440 is a local version identifier. 
Development versions in PEP440 are named `1.0.0betaX` or `1.0.0alphaX`.

So supposing that you have your `plugin.video.foo` in gotham (py3 compatible) with version 1.0.0:

version in gotham is `1.0.0`
version in matrix is `1.0.0+matrix.1`
version for development (external repo) is `1.0.1beta1`

`1.0.1beta1` > `1.0.0+matrix.1` > `1.0.0`

kodi_addon_checker/versions.py
enen92: Core tests only include `beta` and `alpha` so I'm honestly not sure what is supported or not. Note that this validates what we allow on the repo submissions  (and consider best practice) and not exactly what is theoretically possible. If some other edge case arrives, we can always improve on what we have now

tests/test_versions.py
enen92: This is a good reminder I need to update the wiki when this versioning stuff is finished

enen92:
(closed and reopen to trigger travis checks against our repo)

PR #227, opened 2020-04-09
Check if kodi migration is possible

enen92:
This PR implements a "forward" version check that complements the "backwards" version check we already had. 

Right now we only check if the addon being submitted has a higher or equal version than the below branches (let's say version for leia has to be greater or equal than the version for gotham). This allows for users to receive the update (if higher) when migrating from gotham to leia.

However, we don't check for the same addon in upper branches. This means that if an addon is submitted to gotham with a version higher than leia, updates will break.

All this does is to do both checks:

- check that in lower branches an addon with the same id has lower or equal versions than the addon I am submitting
- check that in upper branches there is no addon with the same id but with version lower than the one I am submitting to a lower branch


enen92:
Agrr, finally ready.

kodi_addon_checker/check_addon_branches.py
```diff
@@ -0,0 +1,116 @@
...
+    is_new_addon = True
+    for branch, repo in sorted(all_repo_addons.items(), reverse=True):
+
+        # Addon submission must be higher than the versions already available in lower
+        # branches to allow users to receive the update
+        if KodiVersion(branch) <= kodi_version and addon_name in repo:
+            is_new_addon = False
+            _check_version_higher(report, addon_details, branch, repo.find(addon_name).version, pr)
+
+        # Addon submission must be lower than the versions already available in upper branches
```
Razzeee, line 41: Did this change? Normally it worked to push a higher version to a lower repo and have it overwrite all repos above too? Or is that due to the py3 changes?

Razzeee: And yes, I know this is confusing for the repo maintainers.

enen92: Yes that is the main change (althought there are others mainly code simplification). This isn't only related to py3. If you update a version for your addon in a lower branch (let's say gotham) and have the same add-on on an upper branch (eg Leia), the version you assign to the add-on in Leia must be lower than the version of the add-on available in Leia. If they are not, a user migrating from gotham to Leia will not receive the update (lower version). If an add-on exist in both repos, theoretically they are not compatible (python api's, language stuff, etc)

enen92: I think you are right and this needs further work

enen92:
@Razzeee you were right, thanks. There's no problem in having a higher version in a lower branch and have it override a lower version in a higher repository (e.g. leia). The problem only arises when the upper branch in fact corresponds to a kodi version that introduces a breaking change in the python abi (eg. matrix).
I adapted the code to only trigger an error if there's no backwards compatibility between python abis and the version in the upper branch is lower than the lower branch - thus preventing a seamless user migration between versions.

Thanks!

kodi_addon_checker/check_addon_branches.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # Addon submission must be lower than the versions already available in upper repo branches
+        # if that branch corresponds to a breaking change (e.g. version of addon in matrix version >
+        # version of addon in gotham) since xbmc.python is not backwards compatible.
+        elif KodiVersion(branch) > kodi_version and addon_name in repo and \
+            not _is_pythonabi_compatible(kodi_version, branch):
+            is_new_addon = False
+            _check_version_lower(report, addon_details, branch, repo.find(addon_name).version, pr)
+
+    if is_new_addon:
```
Razzeee, line 50: I think you might now end up here for some cases that are not new addons?

enen92: I was about to remove the `is_new_addon = False` from the above condition. The simple fact the addon is in a lower repository makes it not a new addon. So I think this is fine as is.

dagwieers:
We get errors on Krypton and Leia:
```
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.2/x64/bin/kodi-addon-checker", line 11, in <module>
    load_entry_point('kodi-addon-checker==0.0.17', 'console_scripts', 'kodi-addon-checker')()
  File "/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/kodi_addon_checker/__main__.py", line 92, in main
    report.add(check_artifact(directory, args, all_repo_addons))
  File "/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/kodi_addon_checker/__main__.py", line 55, in check_artifact
    return check_addon.start(artifact_path, args, all_repo_addons, config)
  File "/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/kodi_addon_checker/check_addon.py", line 50, in start
    check_addon_branches.check_for_existing_addon(addon_report, addon_path, all_repo_addons, args.PR,
  File "/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/kodi_addon_checker/check_addon_branches.py", line 46, in check_for_existing_addon
    not _is_pythonabi_compatible(kodi_version, branch):
  File "/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/kodi_addon_checker/check_addon_branches.py", line 70, in _is_pythonabi_compatible
    VERSION_ATTRB['xbmc.python'][target_branch]["min_compatible"]
TypeError: unhashable type: 'KodiVersion'
```

dagwieers:
This puts us in a bind as we need fixes from the master branch, but the latest released breaks our  automated testing in GitHub CI.

PR #228, opened 2020-04-09
Added check for duplicate context menu libraries/args

basrieter:
I added an option to validate that the context menu items in the addon.xml do not define duplicate `library` parameters (for Leia and down) or `library`/`args` combinations for Matrix (See https://github.com/xbmc/xbmc/pull/17631)

### DIFF
--- a/kodi_addon_checker/xml_schema/matrix_contextitem.xsd
+++ b/kodi_addon_checker/xml_schema/matrix_contextitem.xsd
@@ -8,6 +8,7 @@
           <xs:unique name="oneLibraryOnly">
             <xs:selector xpath="./*"/>
             <xs:field xpath="@library"/>
+            <xs:field xpath="@args"/>
           </xs:unique>
         </xs:element>
       </xs:sequence>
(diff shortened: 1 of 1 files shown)


kodi_addon_checker/xml_schema/jarvis_contextitem.xsd
```diff
@@ -4,7 +4,12 @@
   <xs:element name="extension">
     <xs:complexType>
       <xs:sequence>
-        <xs:element name="menu" type="menuType" minOccurs="1" maxOccurs="unbounded"/>
+        <xs:element name="menu" type="menuType" minOccurs="1" maxOccurs="unbounded">
```
enen92, line 7: context menu items were introduced in isengard so I assume you need a similar check on other schemas too

basrieter: I see, let me fix that.

basrieter: From `isengard_contextitem.xsd` it seems that the nesting is done by using multiple `extension` nodes. As the XSD only operates on a single Node, we won't be able to validate it with an XSD.

So my suggestion would be to go forward with this PR and let Isengard be as it is.

PR #230, opened 2020-04-11
Improve language folder layout check for versions below isengard

enen92:
Currently we only check if in versions higher or equal to isengard the addon includes the new language folder layout. However, we don't to anything for version between gotham and isengard. If an addon update sends a "new folder layout" for gotham and helix, kodi won't be able to use them, so error out.

### DIFF
--- a/script.test/main.py
+++ b/script.test/main.py
@@ -108,7 +108,9 @@ def start(addon_path, repo_addons, config=None):
             _check_for_legacy_strings_xml(addon_report, addon_path)
 
             if branch_name not in ['gotham', 'helix']:
-                _check_for_legacy_language_path(addon_report, addon_path)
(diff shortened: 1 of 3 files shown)


script.test/main.py
```diff
@@ -108,7 +108,9 @@ def start(addon_path, repo_addons, config=None):
             _check_for_legacy_strings_xml(addon_report, addon_path)
```
Razzeee, line 108: We should probably move away from the idea, that the test script needs to be in sync or even is the same python source. I would prefer it to be just some random addon.

enen92: Yes, I also don't like that kind of code duplication. We probably want to test this with a bunch of addons having known problems.

PR #234, opened 2020-04-28
Improve forward-version/migration checks

enen92:
Unfortunately the last approach do not work for all addons, causing problems in the repo-scrappers repository. The repository generator places clones of the same addon on upper branches if their dependencies are abi backwards compatible.
Means that if you push a xml scrapper to gotham branch, the same version will also be available in matrix branch (I mean in the addons.xml.gz and not in github). Addon-checker errors out because it sees the matrix addon as non-migration compatible (same addon version). This happens because we are blindly making that decision on the fact `xbmc.python` is not forward compatible between target and matrix - but the addon might not depend on xbmc.python at all.

So this PR changes the check to make it more generic, it looks over all the xbmc api dependencies (xbmc.python, xbmc.gui, xbmc.metadata, etc) and checks if they are forward compatible between the repo to which the addon is being submitted and the "upper branch" that also contains the addon (and just if the addon actually depend on any of them).

### DIFF
--- a/kodi_addon_checker/check_addon_branches.py
+++ b/kodi_addon_checker/check_addon_branches.py
@@ -7,9 +7,9 @@
 """
 
 import logging
-import os
 import xml.etree.ElementTree as ET
 
+from .addons.Addon import Addon
 from .check_dependencies import VERSION_ATTRB
 from .record import INFORMATION, PROBLEM, Record, WARNING
 from .report import Report
@@ -19,16 +19,21 @@ from .versions import AddonVersion, KodiVersion
 LOGGER = logging.getLogger(__name__)
 
 
-def check_for_existing_addon(report: Report, addon_path: str, all_repo_addons: dict, pr: bool,
-                             kodi_version: KodiVersion):
+def check_for_existing_addon(report: Report, addon: Addon, all_repo_addons: dict, args):
     """Check if addon submitted already exists or not
-        :addon_path: path of the addon
+        :report: the report object
+        :addon: the Addon object (contains id and version)
         :all_repo_addons: dictionary return by all_repo_addon() function
+        :args: the args object passed to addon-checker
     """
-
-    addon_xml = os.path.join(addon_path, "addon.xml")
-    addon_name, addon_version = _get_addon_name(addon_xml)
+    # addon details
+    addon_name = addon.id
+    addon_version = addon.version
     addon_details = {'name': addon_name, 'version': addon_version}
+
+    # args
+    pr = bool(args.PR)
+    kodi_version = KodiVersion(args.branch)
 
     is_new_addon = True
     for branch, repo in sorted(all_repo_addons.items(), reverse=True):
@@ -41,9 +46,9 @@ def check_for_existing_addon(report: Report, addon_path: str, all_repo_addons: d
 
         # Addon submission must be lower than the versions already available in upper repo branches
         # if that branch corresponds to a breaking change (e.g. version of addon in matrix version >
-        # version of addon in gotham) since xbmc.python is not backwards compatible.
+        # version of addon in gotham) since there might be dependencies not abi-backward compatible.
         elif KodiVersion(branch) > kodi_version and addon_name in repo and \
-            not _is_pythonabi_compatible(repr(kodi_version), branch):
+            not _is_xbmcabi_compatible(addon.dependencies, repr(kodi_version), branch):
             is_new_addon = False
             _check_version_lower(report, addon_details, branch, repo.find(addon_name).version, pr)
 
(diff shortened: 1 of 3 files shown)


kodi_addon_checker/check_addon_branches.py
```diff
@@ -19,17 +19,20 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def check_for_existing_addon(report: Report, addon_path: str, all_repo_addons: dict, pr: bool,
-                             kodi_version: KodiVersion):
+def check_for_existing_addon(report: Report, addon: Addon, all_repo_addons: dict, args):
     """Check if addon submitted already exists or not
         :addon_path: path of the addon
```
Razzeee, line 24: Can you align the docs?

kodi_addon_checker/check_addon_branches.py
```diff
@@ -60,14 +63,20 @@ def _get_addon_name(xml_path: str):
     return (tree.get("id"), tree.get("version"))
 
 
-def _is_pythonabi_compatible(target_branch, upper_branch):
+def _is_xbmcabi_compatible(dependencies, target_branch, upper_branch):
```
Razzeee, line 66: Would prefer these to be typed, but I'm not really a python pro :)

kodi_addon_checker/check_dependencies.py
```diff
@@ -56,7 +56,43 @@
         'krypton': {'min_compatible': '2.1.0', 'advised': '2.25.0'},
         'leia': {'min_compatible': '2.1.0', 'advised': '2.26.0'},
         'matrix': {'min_compatible': '3.0.0', 'advised': '3.0.0'}
-    }
+    },
+    'xbmc.gui': {
+        'gotham': {'min_compatible': '5.0.0', 'advised': '5.0.0'},
```
Razzeee, line 61: Can you check if it would make sense to default to `AddonVersion` objects here? So that we don't need to parse these for every useage?

kodi_addon_checker/check_addon_branches.py
enen92: I don't see them misaligned, the description is just small and the last arg description is a bit bigger. Is this what you mean?

kodi_addon_checker/check_addon_branches.py
enen92: will add

kodi_addon_checker/check_dependencies.py
enen92: you can disable checks I suppose. Hence this may not even be used. I think it's a good a idea but out of the scope of this PR

kodi_addon_checker/check_addon_branches.py
Razzeee: `addon` is missing. `addon_path` -> `report` and needs explanation. Not sure if `args` should get any docs.

Razzeee:
Should we add the scrapers repo to the tests?

Razzeee:
This seems buggy. https://travis-ci.org/github/xbmc/addon-check/jobs/681138217 errors for `script.module.actionhandler` but I'm not seeing the problem it suggests in the repo. https://github.com/xbmc/repo-scripts/tree/matrix

Not sure if I just misunderstood.

enen92:
That's odd, was it the only one you found? For some reason the addons.xml.gz from the repo lists that script.module in matrix?

enen92:
I see the problem: https://mirrors.kodi.tv/addons/matrix/script.module.actionhandler/

For some reason this addon doesn't have any xbmc.python dependency:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<addon id="script.module.actionhandler" name="ActionHandler" version="1.0.3" provider-name="Philipp Temminghoff (phil65)">
	<extension point="xbmc.python.module" library="lib" />
	<extension point="xbmc.addon.metadata">
		<summary lang="en">Helper module for Event handling in WindowXMLs</summary>
		<description lang="en">Helper module to allow mapping Actions, Onfocus and click events via decorators</description>
		<platform>all</platform>
		<license>GNU GENERAL PUBLIC LICENSE Version 2.1, February 1999</license>
		<source>https://github.com/phil65/script.module.actionhandler</source>
	</extension>
</addon>
```

The problem is that the addon in jarvis doesn't have any dependency on xbmc.python and that made it propagate all the way till matrix:

https://github.com/xbmc/repo-scripts/blob/jarvis/script.module.actionhandler/addon.xml

This should be fixed, but it's not related to this PR

enen92:
Repo-scrappers should be added to the tests since rechi configured it there

Razzeee:
> That's odd, was it the only one you found?

I basically clicked on random through some of the test runs and only found that. There might be more.

enen92:
I'll take a look over the logs to find more references of `forward abi compatible dependencies` tomorrow. Although I think it will actually help us find similar problems to the one above

Razzeee:
Agreed, still good to make sure there is no wrong behavior that we can see.

enen92:
@Razzeee looked into the logs and found 2 addons broken in matrix (pr's to solve it are above ^^). Problem is that those addons did not define `xbmc.python` as a requirement and that made them being included in the matrix branch. I'd say this is good to go.

PR #236, opened 2020-05-02
Ignore the log files when they are generated

### LINKED ISSUE #233: Log file generates a warning when --reporter=log is used
When I run `kodi-addon-checker --branch leia --reporter=log` there is a warning reported on the log file generated by the addon-checker:
```
WARN: Found non whitelisted file ending in filename ./plugin.video.imagotv/kodi-addon-checker-report.log
WARN: We found no problems and 1 warnings, please check the logfile.
```

I think it is wrong and this file should somehow be ignored.
It seems it was already reported in #156 but from what I see in the [final answer](https://github.com/xbmc/addon-check/issues/156#issuecomment-462079242) the reporter closed the issue because he/she didn't use the argument `reporter=log`anymore which doesn't really fix the initial issue.

I didn't look in the code yet so I don't know how easy it could be fixed but could you first confirm if it is a bug or if it is somehow expected?

Thanks :)

thombet:
Fixes #233 

* Create two functions in the common module to get the full path of the
log files (debug log and reporter log) anywhere in the script. It
avoids duplicate reference to the names of the log files to keep the code DRY.
* When a log file is generated it will be whitelisted to avoid reporting warnings.
* Files with the same name as the log files but not in the current folder will still be reported (these files were generated from a previous run of the checker so they should not be ignored)


thombet:
Some additional information:
* It seems both `.format()` and `%` are used in the code but I used `%` in `__main___.py` to be consistent with the rest of this file.
* From a design point of view I think it would have been better to define `get_reporter_log_path()` in the `LogReporter` class (plugins/log_reporter.py) but I didn't find a good way to use this class accross the code since it is imported with `load_plugins()` in `__main__.py`. If you have a better idea, feel free to suggest it :)
* I ran several (manual) tests locally to confirm it is working as expected but should I create any auto tests?

thombet:
The tests failed because the stub of argParser does not work anymore with my change.
I will fix this and probably add some tests to reproduce the use cases I tested manually.

enen92:
There are a lot of use cases for use with this tool, namely via pipelines. I am not sure I like the idea of ignoring specific files that may end up bundled with the addon submission.
Maybe a possible option is to only allow usage of --log or --debug log if an additional argument (log path) is submitted? 
Maybe also error out if the submitted log path is equal to the directory we are checking for validation?

Another option is to make it platform dependent and make it log by default to the standard log location per platform (/var/log/addon-check on linux, ~/Library/Logs/ on osx, etc) and at the end of the execution state the log can be found in xxxx ?

thombet:
I came up with this solution because I thought requesting a path for the log files would break the compatibility with previous versions but your idea of using a default path solves this 🙂 

I agree this PR adds a risk of submitting ignored files when used in pipelines so I will push a new update implementing your ideas. Thank you for suggesting them @enen92 👍

Razzeee:
Should we close this until you find time to work on this again?

thombet:
> Should we close this until you find time to work on this again?

I can't believe it's been 4 months already...
Yes I will close this pull request and I will upload a new one later, thanks for the reminder.

PR #239, opened 2020-07-05
fix any Gettext automatic comments in the source po before testing with polib

### LINKED ISSUE #237: Checker failing for Gettext automatic comments
See https://github.com/xbmc/repo-scripts/pull/1561

anxdpanic:
Believe this resolve #237


anxdpanic:
Looks like I introduced a syntax error in the process, should be an easy fixup.

anxdpanic:
Should be good to go now

enen92:
I'd prefer the problem to be fixed in polib but this is also an option. Thanks much for fixing this!

PR #240, opened 2020-08-23
Add "lifecyclestate" check to addon.xml

AlwinEsch:
This changes related to https://github.com/xbmc/xbmc/pull/18286 to allow a addon.xml with new:
- `<lifecyclestate type="broken" lang="en_GB">SOME TEXT</lifecyclestate>`

Further added test for binary addons with them and except screensaver works all with them. By screensaver already the python one inside, maybe it make sense to use "library" there as optional?

Also has not found a way to make with binary minimal one if the `library_...` as required, 
Thought to use `xs:assert` but then it required `vc:minVersion="1.1"` and my experience with it is to small to change for.


kodi_addon_checker/xml_schema/matrix_binary_audiodecoder.xsd
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE schema PUBLIC "-//W3C//DTD XMLSCHEMA 200102//EN" "http://www.w3.org/2001/XMLSchema.dtd">
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:element name="extension">
+    <xs:complexType>
+      <xs:attribute name="point" type="xs:string" use="required"/>
```
enen92, line 6: Doesn't this have a fixed value for this type of add-on? Same applies to all the other new schemas

kodi_addon_checker/xml_schema/matrix_binary_audiodecoder.xsd
```diff
@@ -0,0 +1,22 @@
...
+      <xs:attribute name="name" type="xs:string"/>
+      <xs:attribute name="extension" type="xs:string"/>
+      <xs:attribute name="tags" type="xs:boolean"/>
+      <xs:attribute name="library_android" type="xs:string"/>
```
enen92, line 10: You can probably restrict this to *.so with a regex pattern whenever applied, dylib for osx and dll in windows. Maybe too much effort as we are not checking binary add-ons at the moment

kodi_addon_checker/xml_schema/matrix_binary_audiodecoder.xsd
AlwinEsch: Seems no, there comes without:
```
failed validating {'point': 'xbmc.player.musicviz', 'library_linux': 'visualization.starburst.so.2.2.0'} with XsdAttributeGroup(['library_android', 'library_freebsd', 'library_rbpi', 'library_linux', 'library_windx', 'library_windows', 'library_windowsstore', 'library_darwin_embedded', 'library_osx', 'library']):

Reason: 'point' attribute not allowed for element.

Schema:

  <xs:complexType xmlns:xs="http://www.w3.org/2001/XMLSchema">

    <xs:attribute name="library_android" type="xs:string" />
    <xs:attribute name="library_freebsd" type="xs:string" />
    <xs:attribute name="library_rbpi" type="xs:string" />
    <xs:attribute name="library_linux" type="xs:string" />
    <xs:attribute name="library_windx" type="xs:string" />
    <xs:attribute name="library_windows" type="xs:string" />
    <xs:attribute name="library_windowsstore" type="xs:string" />
    <xs:attribute name="library_darwin_embedded" type="xs:string" />
    <xs:attribute name="library_osx" type="xs:string" />
    <xs:attribute name="library" type="xs:string" />
  </xs:complexType>

Instance:

  <extension point="xbmc.player.musicviz" library_linux="visualization.starburst.so.2.2.0" />

Path: /extension
```

kodi_addon_checker/xml_schema/matrix_binary_audiodecoder.xsd
AlwinEsch: I can look about should be possible by regex.

kodi_addon_checker/xml_schema/matrix_binary_audiodecoder.xsd
AlwinEsch: Or you mean to check about name, seen it nowhere on all files and think becomes done by this https://github.com/xbmc/addon-check/blob/master/kodi_addon_checker/schema_validation.py#L50, as without correct name it comes never to the file :thinking:

enen92: Ah you're right, we validate this one in the code and use it to match to the schema

kodi_addon_checker/xml_schema/matrix_metadata.xsd
```diff
@@ -0,0 +1,109 @@
...
+        <xs:element name="description" type="translatedStringLong" minOccurs="0" maxOccurs="unbounded"/>
+        <xs:element name="disclaimer" type="translatedString" minOccurs="0" maxOccurs="unbounded"/>
+        <xs:element name="platform" type="platformList" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="language" type="xs:string"  minOccurs="0" maxOccurs="1"/>
+        <xs:element name="license" type="nonEmptyString"  minOccurs="0" maxOccurs="1" />
+        <xs:element name="forum" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="website" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="source" type="nonEmptyString" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="email" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        <xs:element name="broken" type="nonEmptyString" minOccurs="0" maxOccurs="1"/> <!-- deprecated and removed in future, done by lifecyclestate -->
```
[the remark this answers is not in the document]
AlwinEsch: Is removed

kodi_addon_checker/xml_schema/matrix_binary_audiodecoder.xsd
AlwinEsch: Have added to all binary parts a regex about, the test at home has worked.

OK:
- audiodecoder.2sf.so.12.2.0
- audiodecoder.2sf.so.12
- audiodecoder.2sf.so
- audiodecoder.blabala.aaa.2sf.so.12.2.0

ERROR (as wanted):
- audiodecoder.2sf.so.12.2
- audiodecoder.2sf.so.12.2a.0
- audiodecoder.2sf.so.12z.2.0
- audiodecoder.2sf.so.
- audiodecoders.2sf.so
- audiodecoder.2sf.sol.12.2.0

enen92:
@alwinesch can you fix the issues brought by the linter? Should be quick (line too long, super, etc).
This should be merged so we can bump the version in pipy

AlwinEsch:
Have it updated, about the `R1725: Consider using Python 3 style super() without arguments (super-with-arguments)` not found a error source and seems after fix of others away.

enen92:
Thanks a lot

Issue #241, opened 2020-09-06
I can not run addon-check due to script.module.pycryptodome


CastagnaIT:
If i try to running addon-check on matrix branch
it complain of script.module.pycryptodome

> INFO: This is a new addon
> INFO: Valid XML file found
> WARN: 405 Client Error: Method Not Allowed for url: https://www.netflix.com/
> ERROR: Required dependency script.module.pycryptodome is not available in current repository
> INFO: Image icon exists
> INFO: icon dimensions are fine 512x512
> INFO: Image fanart exists
> INFO: fanart dimensions are fine 1920x1080

Flow: https://github.com/CastagnaIT/plugin.video.netflix/runs/1077533279

I have see that there is this ignore for Leia, should be added also for Matrix? is this the cause?
https://github.com/xbmc/addon-check/blob/master/kodi_addon_checker/check_dependencies.py#L200

PR #242, opened 2020-09-06
Ignore script.module.pycryptodome dependency from Leia and later versions

CastagnaIT:
I have changed the ignore script.module.pycryptodome dependency to include all releases from Leia and later versions
only after this i can successfully get right result in the test buildings flow

change tested locally, flow: https://github.com/CastagnaIT/plugin.video.netflix/runs/1077732346

Ref Issue: #241

### DIFF
--- a/kodi_addon_checker/check_dependencies.py
+++ b/kodi_addon_checker/check_dependencies.py
@@ -197,7 +197,7 @@ def _get_ignore_list(kodi_version: KodiVersion):
        on the branch name
     """
 
-    if kodi_version == KodiVersion("leia"):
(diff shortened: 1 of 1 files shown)


CastagnaIT:
i would sugget to change this line:

`if kodi_version in [KodiVersion("leia"), KodiVersion("matrix")]:`

in to:

`if kodi_version != KodiVersion("krypton"):`

to not make this change on each new version of kodi
what do you think?

CastagnaIT:
can not be merged?

CastagnaIT:
@Razzeee can you include this pr in new release?

Razzeee:
Thank you

PR #243, opened 2020-09-06
[linting] use Python 3 style super() without arguments (pylint=super-with-arguments)

anxdpanic:


CastagnaIT:
after this still works on python 2?

Razzeee:
As far as I remember this always targeted python3 as environment, but it parsed both py2 and py3 addons.

CastagnaIT:
yes i misunderstood sorry ..bad day

anxdpanic:
Yea, this is to resolve travis issues in #242

CastagnaIT:
I do not understand why you say this:

> Yea, this is to resolve travis issues in #242

currently is not solved:
https://github.com/CastagnaIT/plugin.video.netflix/runs/1101050382

The problem is the dependency,
from what i see, can be solved only with my PR

anxdpanic:
@CastagnaIT 
It was literally for your pull request (and future ones), Travis failed on your PR due to this reason. I just restarted the test and you can see https://travis-ci.org/github/xbmc/addon-check/builds/724630716 that the 4 required tests are passing now.

CastagnaIT:
ah in this sense
i will rebase my branch

PR #247, opened 2021-02-12
Dependencies: Always consider the higher dependency version in the re…

enen92:
…po if multiple copies exist for the same addon id

This fixes https://github.com/xbmc/repo-plugins/pull/3350


anxdpanic (approved):

Razzeee (commented):
LGTM

PR #249, opened 2021-02-17
warn when no po files are available

anxdpanic:


kodi_addon_checker/check_string.py
```diff
@@ -66,8 +66,11 @@ def check_for_invalid_strings_po(report: Report, file_index: list, addon_path: s
             report_made = True
 
     if not en_gb_present:
-        report_made = True
-        report.add(Record(PROBLEM, "Required default language 'en_gb' is not present."))
+        if po_file_index:
+            report_made = True
+            report.add(Record(PROBLEM, "Required default language 'en_gb' is not present."))
+        else:
+            report.add(Record(WARNING, "No PO files found, consider translating."))
```
enen92, line 73: If there is no po files available it doesn't mean there is something to translate. Maybe make this more generic?  Something in the lines of "Consider adding language support to your add-on"

anxdpanic: Makes sense, updated.

enen92 (approved):

PR #252, opened 2021-03-20
Preparations for master branch rename

razzeee:


enen92:
We've decided a while ago in some team meeting to keep using the master branch naming as long as it's allowed. All this main rename is just non-sense IMOH.

PR #254, opened 2021-04-12
Report the problem if the addon.xml xml contents are invalid

enen92:
Addon checker is throwing if invalid xml data is present in addon xml.
Catch the exception and log a problem instead.


sualfred:
LGTM.

PR #255, opened 2021-04-12
Check if addon version is compliant with repository generator

enen92:
Follow up to https://github.com/xbmc/repository-generator/pull/16, this makes addon-checker validate the provided version in addon.xml against the regex used in repository-generator. This makes sure the checker catches any invalid version in advance

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ It can also be used locally for detecting problems in your addons.
 
 - Checks if addon.xml and license file exists for an addon.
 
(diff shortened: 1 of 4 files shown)


yol:
Can you add `2.3.0-backported-Leia` (or similar) as check that must fail? Otherwise looks good :+1:

enen92:
Done, I'll merge if travis is happy

yol (approved):

PR #262, opened 2023-02-19
Add Omega branches

fuzzard:
Hope this is ok @enen92 . let me know if ive missed something, or just flat out did something wrong.

### DIFF
--- a/.github/workflows/check-kodi-repos.yml
+++ b/.github/workflows/check-kodi-repos.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         repo: ['repo-plugins', 'repo-scripts', 'repo-scrapers', 'repo-resources', 'repo-skins']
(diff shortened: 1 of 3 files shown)


enen92 (approved):
Yes looks correct (my fault on the add-on API for nexus). Note that we usually don't accept addons into Dev branches until pretty close to release (and sometimes even only after release). 
There's no advantage currently for people to submit to omega

enen92:
On a sidenote can you also update ghactions to look into omega as well?
https://github.com/xbmc/addon-check/blob/master/.github/workflows/check-kodi-repos.yml#L15

enen92:
Also, tests on the CI are failing because the repo-generator is not yet generating the gz file for omega: 
http://mirrors.kodi.tv/addons/omega/addons.xml.gz

I guess that PR needs to be merged before considering this one

fuzzard:
Updated the check-kodi-repos.yml

enen92:
Actions are green, merging

PR #264, opened 2023-03-25
Add option `skip-dependency-checks`

### LINKED ISSUE #263: Add option to ignore dependency checks
I run `addon-check` via Github actions in my plugins Github repositories.
When I update the required versions for my dependencies and they are not published yet (in the official Kodi repository), it always fails with this kind of error message:
```
ERROR: Version mismatch for required dependency script.module.xyz, required: 2.2.3, Available: 2.2.2
```
So it would be nice to have an option to ignore dependency checks (at least for my use-case).

goggle:
Adds an option to ignore dependency checks.
The reason for this is outlined in #263.
Closes #263.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,4 +101,5 @@ You can use the tool with the following options:
 --allow-folder-id-mismatch  allow the addon's folder name and id to mismatch
 --reporter                  enable a reporter, this option can be used multiple times
 --enable-debug-log          enable debug logging to kodi-addon-checker.log
+--skip-dependency-checks    do not check if addon dependencies are available in the official Kodi addon repository
 ```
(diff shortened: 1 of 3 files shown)


razzeee (approved):

razzeee:
I still think we should go for modes and not for single settings, like outlined in https://github.com/xbmc/addon-check/pull/158#issuecomment-461417210

goggle:
Are the failing `GitHub Action` checks my fault? Have I forgotten to add something?

razzeee:
Seems like the tests expect `skip_dependency_checks` to be there now. Either the default is not working or it's not affecting the test suite (as it might only start below that layer)

romanvm:
@razzeee is this ok to merge?

razzeee:
Seems straight forward enough

PR #266, opened 2023-03-27
Remove travis file

razzeee:
This might be the cause of the CI hanging


razzeee:
Doesn't seem to solve it, might need to readd and delete travis from their side? I don't see any other mention of it in our code, other then the readme.

We also need to remove it from the readme

PR #271, opened 2024-05-14
Rate limit requests for addons.xml.gz

MoojMidge:
Rate limiting has been added to mirrors.kodi.tv which the multiple requests for the various repository addons.xml.gz will trigger.

This adds some rate limiting, error handling and retry backoff to avoid the rate limit resulting in the addon checker failing

### DIFF
--- a/kodi_addon_checker/addons/Repository.py
+++ b/kodi_addon_checker/addons/Repository.py
@@ -24,7 +24,7 @@ class RateLimitedAdapter(requests.adapters.HTTPAdapter):
         self._wait_time = wait
         max_retries = requests.adapters.Retry(
             total=retries,
-            backoff_factor=wait,
+            backoff_factor=wait or 10,
             status_forcelist={429, },
             allowed_methods=None,
         )
@@ -32,7 +32,7 @@ class RateLimitedAdapter(requests.adapters.HTTPAdapter):
         super().__init__(*args, **kwargs)
 
     def send(self, *args, **kwargs):
-        if self._last_send:
+        if self._wait_time and self._last_send:
             delta = time.time() - self._last_send
             if delta < self._wait_time:
                 time.sleep(self._wait_time - delta)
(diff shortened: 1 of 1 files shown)


razzeee (approved):

kodi_addon_checker/addons/Repository.py
```diff
@@ -16,26 +18,59 @@
...
+    _adapter = RateLimitedAdapter(retries=5, wait=10)
+    _session.mount('http://', _adapter)
+    _session.mount('https://', _adapter)
+    atexit.register(_session.close)
+
     def __init__(self, version, path):
         super().__init__()
         self.version = version
         self.path = path
+        self.addons = []
```
MoojMidge, line 60: Maybe this line should be moved back to where it was, so that the checker will fail if the versioned repository download fails due to some other server error that cannot be resolved by the allowed retries?

MoojMidge:
One issue with this PR is that, while it does prevent the rate limiting, it may allow for the possibility of the checker appearing to  complete successfully, if some other server error occurs.

In this instance the checker would just run with an empty list of addons for the versioned repository that failed to download. Perhaps this should still raise an exception?

It was also suggested that not all these repositories (currently Gotham and up) should be checked anymore. Maybe only Leia and above?

basrieter:
Could we not just send a specific header that will disable the rate limit for that call? The main reason we limit the rate is the normal repository fetching of addon.xml(.md5) that floods the servers. So we could perhaps differentiate on the agent header in the rate limiter?

MoojMidge:
Yes, that is also an option, but would only apply to mirrors.kodi.tv itself.

The actual mirrors themselves would not respond in the same manner.

I updated the PR:
- removed the explicit time delays between requests, while still keeping the retry back-off
- instead limit the number of connections per host to 3
- applies to mirrors.kodi.tv and any mirrors used, up to the default connection pool limit of 10
- will block any further connections per host until the existing 3 connections to the host have finished

Should be a little bit more friendly to all the servers involved.

Also moved the `self.addons = []` line back to where it was originally, so that the checker would fail if the list of addons still failed to download for some reason.

basrieter (approved):

razzeee:
Thanks for working on this

MoojMidge:
Just realised there is an issue with this PR if used with older distributions that have old versions of `requests`/`urllib3` e.g. Ubuntu 20.04

`requests.adapters.Retry` is using the `allowed_methods` parameter, which was introduced in [v1.26 of urllib3](https://urllib3.readthedocs.io/en/latest/changelog.html#id33) a bit over 3 years ago.

Shall I create a PR for https://github.com/xbmc/addon-check/blob/master/requirements.txt to use this as the minimum version required?

basrieter:
Thay makes sense.

PR #273, opened 2024-05-28
Update min required version of requests and urllib3

MoojMidge:
Refer https://github.com/xbmc/addon-check/pull/271#issuecomment-2132440128

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 packaging
 pillow
 polib
-requests
+requests >= 2.25.0
 radon
-urllib3
+urllib3 >= 1.26.0
 xmlschema


MoojMidge:
The addon check failure indicates that the previous PR may not be enough to.fully eliminate the download errors from the mirrors, if multiple checks happen at the same time

https://github.com/xbmc/addon-check/actions/runs/9262295587/job/25478978358?pr=273

Maybe further reduce the number of allowed connections per host to 1?

basrieter (approved):

basrieter:
> The addon check failure indicates that the previous PR may not be enough to.fully eliminate the download errors from the mirrors, if multiple checks happen at the same time
> 
> https://github.com/xbmc/addon-check/actions/runs/9262295587/job/25478978358?pr=273
> 
> Maybe further reduce the number of allowed connections per host to 1?

For now, it seems to do a good job. Let's see how it goes.

PR #283, opened 2025-05-10
Fix checking language addons

romanvm:
Do not fail checking language resource addons because of a missing en-gb .po file.


razzeee:
Can you rebase this?

